PR #39, opened 2020-09-20
Feature/opsani dev enhancements

blakewatters:


blakewatters:
Replaced with rebased flavor #44

PR #48, opened 2020-10-06
Feature/newrelic connector

blakewatters:


servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def __check__(self) -> Check:
+        return Check(
+            name=f"Check {self.name}",
+            description=f"Run Newrelic get \"{self.fetch_name}: {self.values_selector}\""
+        ) # TODO Checker class
+
+# TODO: what is this used for?
+# class NewrelicTarget():
+#     pass # TODO
```
blakewatters, line 72: I am guessing this is lifted from Prometheus. Targets in Prometheus refer to the instances being scraped. There's no equivalent in New Relic and this can be dropped.

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+    """The resolution of the metrics to be fetched.
+    
+    The step resolution determines the number of data points captured across a
+    query range.
+    """
+
+    # targets: Optional[List[NewrelicTarget]] TODO: what is this used for?
+    """An optional set of Newrelic target descriptors that are expected to be
+    scraped by the Newrelic instance being queried.
+    """
```
blakewatters, line 102: Can be removed

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+                    values_selector="requests_per_minute"
+                ),
+                NewrelicMetric(
+                    "error_rate", Unit.COUNT, "Errors/all", "error_count"
+                ),
+            ],
+            **kwargs,
+        )
+    
+    @validator("base_url", allow_reuse=True)
```
blakewatters, line 128: Don't need `allow_reuse`, will clean up on Prometheus (assuming it as cargo culted)

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+    base_url: AnyHttpUrl
+    fetches: Dict[str, Set[str]]
+    start: datetime
+    end: datetime
+    step: Duration
+
+    @property
+    def params(self) -> Dict[str, str]:
+        return { 
+            'names[]': '+'.join(self.fetches.keys()), # TODO can these be lists and httpx will extrapolate?
```
blakewatters, line 147: Can't remember offhand. Check https://www.python-httpx.org/api/ or write a test.

`respx` makes this kinda thing easy to test through mocking

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+    """NewrelicConnector objects enable servo assemblies to capture
+    measurements from the [Newrelic](https://newrelic.com/) metrics server.
+    """
+    config: NewrelicConfiguration
+
+    # TODO
+    # @on_event()
+    # async def check(self,
+    #     filter_: Optional[Filter] = None, 
+    #     halt_on: HaltOnFailed = HaltOnFailed.requirement
```
blakewatters, line 176: API changed here a bit in v0.8.0

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+    def metrics(self) -> List[Metric]:
+        """Returns the list of Metrics measured through Newrelic queries.
+
+        Returns:
+            List[Metric]: The list of metrics to be queried.
+        """
+        return self.config.metrics
+
+    @on_event()
+    def instances(self) -> List[str]:
```
blakewatters, line 216: Since that endpoint is freeform and has been replaced in `oco-e`, this isn't the right place for this functionality. I'd expect this to live in another connector like Harness or whatever that is responsible for providing the instances and New Relic connectr would just call out to it rather than trying to provide an implementation itself.

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+        f_names_to_ms: Dict[str, List[NewrelicMetric]] = defaultdict(list)
+        fetches: Dict[str, Set[str]] = defaultdict(set)
+        for m in metrics:
+            f_names_to_ms[m.fetch_name].append(m)
+            fetches[m.fetch_name].add(m.values_selector)
+
+        newrelic_request = NewrelicRequest(base_url=self.config.api_url, fetches=fetches, start=start, end=end, step=self.config.step)
+
+        self.logger.trace(f"Getting Newrelic instance ids: {newrelic_request.url}")
+        instance_ids: List[str] = await self.dispatch_event("instances")
```
blakewatters, line 302: Guessing that this probably should fall back to gathering all instances if nothing responds to the instances event to scope it? That would mean that for simpler deployments we just grab it all and you don't have to worry about instance scoping

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+        opsani_config = response.json()
+        canary = opsani_config.get('adjustment', {}).get('control', {}).get('userdata', {}).get('canary')
+        canary_ip = canary.get('ip')
+        canary_ip = re.sub(r"\.", "-", canary_ip)
+        canary_hostname = f"ip-{canary_ip}.ec2.internal"
+
+        async with httpx.AsyncClient(
+            base_url=self.config.api_url + 'applications/{newrelic_app_id}/instances.json'.format(NEWRELIC_APM_APP_ID), 
+            params={'filter[hostname]': canary_hostname},
+            headers={'X-Api-Key': NEWRELIC_APM_API_KEY},
```
blakewatters, line 234: `NEWRELIC_APM_API_KEY` must come from config to support multiple New Relic connector instances

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+                raise
+
+        opsani_config = response.json()
+        canary = opsani_config.get('adjustment', {}).get('control', {}).get('userdata', {}).get('canary')
+        canary_ip = canary.get('ip')
+        canary_ip = re.sub(r"\.", "-", canary_ip)
+        canary_hostname = f"ip-{canary_ip}.ec2.internal"
+
+        async with httpx.AsyncClient(
+            base_url=self.config.api_url + 'applications/{newrelic_app_id}/instances.json'.format(NEWRELIC_APM_APP_ID),
```
blakewatters, line 232: `NEWRELIC_APM_APP_ID` must come from config to support multiple New Relic connector instances

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+        self.logger.trace(f"Querying Newrelic: {newrelic_request.url}")
+        readings = []
+        # TODO asyncio gather this instead
+        for i in instance_ids:
+            api_path = '/applications/{app_id}/instances/{instance_id}/metrics/data.json'.format(NEWRELIC_APM_APP_ID, i)
+            self.logger.trace(f"Querying Newrelic for instance: {i}")
+            async with httpx.AsyncClient(
+                base_url=self.config.api_url + api_path, 
+                params=newrelic_request.params,
+                headers={'X-Api-Key': NEWRELIC_APM_API_KEY},
```
blakewatters, line 313: `NEWRELIC_APM_API_KEY` must come from config to support multiple New Relic connector instances

blakewatters (commented):
@linkous8 I opened a PR to get the review interface

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+DEFAULT_BASE_URL = 'https://api.newrelic.com'
+API_PATH = "/v2"
+
+# TODO: how are secrets handled?
+NEWRELIC_ACCOUNT_ID = str(open('/run/secrets/optune_newrelic_account_id').read()).strip()
+NEWRELIC_APM_API_KEY = str(open('/run/secrets/optune_newrelic_apm_api_key').read()).strip()
+NEWRELIC_APM_APP_ID = str(open('/run/secrets/optune_newrelic_apm_app_id').read()).strip()
+
```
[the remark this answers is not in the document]
linkous8: Updated config https://github.com/opsani/servox/blob/bc8d66cd934ddac09978541bc8d8471a473bf788/servo/connectors/newrelic.py#L62

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,340 @@
...
+class NewrelicMetric(Metric):
+    """NewrelicMetric objects describe metrics that can be measure from the Newrelic apm api."""    
+    
+    fetch_name: str
+    """The name of the APM metric containing the values for this Metric
```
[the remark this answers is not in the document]
linkous8: But its not a query. Its not even the complete query since each metric also requires the `values_selector` property unlike prometheus whose metrics could function with the singular `query` property.

I would be open to renaming the properties to `name_param` and `values_param` since they're turned into params for the http request here: https://github.com/opsani/servox/blob/bc8d66cd934ddac09978541bc8d8471a473bf788/servo/connectors/newrelic.py#L140

However, I feel like calling it a query is shoehorning the typing

servo/connectors/newrelic.py
linkous8: It API requests to newrelic right here: https://github.com/opsani/servox/blob/bc8d66cd934ddac09978541bc8d8471a473bf788/servo/connectors/newrelic.py#L224

If it weren't for that, I would have put it in a standalone connector

servo/connectors/newrelic.py
linkous8: I'd actually been meaning to ask. Should a toggle be defined in the config as to whether the NR driver should respond the instances event? (for cases where the ec2asg connector needs to respond to it instead)

As for getting all instances, normally instances fetched from NR are filtered on a hostname which must be retrieved from OCO config (where the harness job sets it IIRC). I don't think getting all instances would provide sane behavior outside of testing environments, but am open to implementation suggestions

servo/connectors/newrelic.py
linkous8: Httpx doesn't extrapolate as expected, will remove the TODO:

```
>>> import httpx
>>> request = httpx.Request("GET", "https://example.org", params={'host': ['test1', 'test2']})
>>> request.url
URL('https://example.org?host=test1&host=test2')
>>> 
```

servo/connectors/newrelic.py
blakewatters: All of this business is potentially quite brittle, coupled to EC2 implementation details, and requires some other out of band actor to populate the data in the `/config` endpoint: https://github.com/opsani/servox/blob/bc8d66cd934ddac09978541bc8d8471a473bf788/servo/connectors/newrelic.py#L210-L222

Rather than putting in an explicit switch on the configuration (though now that I think about it allowing a hard-coded `instances` list may be useful), what I was thinking is using the presence of one or more `instances` event handlers to override the default behavior of just grabbing the full instances list. This would let you evict the `/config` and EC2 code smells and switch the behavior based on if there are any results returned from a call to `dispatch_event("instances")`.

While I generally agree with your observation about the relatively limited utility of a default implementation that returns all the instances in today's state of affairs with Intuit and Ancestry, as we take Opsani Dev to market and have more middle-market customers my thinking is that we will start to see more engagements where the customer only has one main application and all the monitoring infrastructure is dedicated to it.

My New Relic API knowledge is a bit rusty, but I also wonder if we can eliminate the need for custom `instances` handlers in 80% of the use cases by supporting optional filtering predicates in the config class. There is an `application_name` attribute in the response payload (see https://docs.newrelic.com/docs/apis/rest-api-v2/get-started/list-application-id-host-id-instance-id#locating_instance_id and https://docs.newrelic.com/docs/apis/rest-api-v2/get-started/list-application-id-host-id-instance-id#locating_instance_id) -- if we supported an exact string matcher and regex pattern, shouldn't that cover most sane implementations you can imagine?

servo/connectors/newrelic.py
blakewatters: See the previous reply

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,362 @@
...
+        async with self.api_client as client:
+            try:
+                response = await client.get('/config')
+                response.raise_for_status()
+            except (httpx.HTTPError, httpcore._exceptions.ReadTimeout, httpcore._exceptions.ConnectError) as error:                
+                self.logger.trace(f"HTTP error encountered during oco config GET {self.optimizer.api_url}/config: {error}")
+                raise
+
+        opsani_config = response.json()
+        canary = opsani_config.get('adjustment', {}).get('control', {}).get('userdata', {}).get('canary')
```
benburdick, line 248: Note that with oco-e, we are using a new "Assets API" to store/retrieve component data with a new url and data structure:

`url = "https://api.opsani.com/accounts/{}/applications/{}/assets/opsani.com/<COMPONENT>".format(
    config["opsani_account"], config["opsani_app_name"]
)`

`canary = opsani_config.get('data')
`

servo/connectors/newrelic.py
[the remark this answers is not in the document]
blakewatters: 👍

servo/connectors/newrelic.py
```diff
@@ -0,0 +1,362 @@
...
+        canary_ip = canary.get('ip')
+        canary_ip = re.sub(r"\.", "-", canary_ip)
+        canary_hostname = f"ip-{canary_ip}.ec2.internal"
```
benburdick, line 251: This substitution only works with a specific customer.  In my more generic testing, these hostnames look more like this: `canary_hostname = f"ip-{canary_ip}.us-west-2.compute.internal"`

This will likely be different per customer, unless they are using AWS defaults, in which case they'll look like my example with the exception of the respective region.  We'll want to make this more dynamic and/or configurable with sane defaults

servo/connectors/newrelic.py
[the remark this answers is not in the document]
blakewatters: I don't love the `_param` suffix because it's conflating the implementation detail of being shoved into an HTTP request with its core meaning.

Couldn't we just eliminate the notion of the fetch and values selector entirely by adopting the metric naming convention that New Relic uses in their docs (e.g., metric.name=`WebFrontend/Queue:average_response_time`) (https://docs.newrelic.com/docs/apis/rest-api-v2/application-examples-v2/average-response-time-examples-v2):

```
Response time = HttpDispatcher:average_call_time + ((WebFrontend/Queue:call_count * WebFrontend/Queue:average_response_time)/ HttpDispatcher:call_count)
```

What is before the colon becomes the metric name and what is after becomes the values. Throw a regular expression constraint onto the field and it flattens out into a single string.

The downside is that you can't batch up requests against a single metric without doing some preprocessing to group them by the source, but I think you are already in that situation because `values_selector` is just a string.

This is something I tried to handle in my sketch last week by decomposing the access into the notion of a query, which provides the source of the data, and the metric which binds a value from a named query to something the servo can report back to the optimizer (https://gist.github.com/blakewatters/310f0df9dd051d7c3584cccb9d8e6370)

servo/connectors/newrelic.py
blakewatters: this whole canary hostname computation feels brittle and inelegant to me, I think we can do better. But in any case, the logic definitely does not belong in the middle of the New Relic connector

servo/connectors/newrelic.py
linkous8: @derezed88 Could we get your thoughts and commentary on the last paragraph of Blake's latest response?

servo/connectors/newrelic.py
linkous8: discussion moved: https://github.com/opsani/servox/pull/48#discussion_r502610994

servo/connectors/newrelic.py
linkous8: @benburdick what does COMPONENT refer to in this case? I could have sworn that was tied to the adjustment and I'm not sure its easily accessed from the measure connector under the current model

servo/connectors/newrelic.py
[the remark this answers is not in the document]
linkous8: See updated implementation: https://github.com/opsani/servox/blob/279ae1fbd2709246baf68615e400c596adf60623/servo/connectors/newrelic.py#L45

servo/connectors/newrelic.py
linkous8: Moved the format string to the configuration: https://github.com/opsani/servox/blob/279ae1fbd2709246baf68615e400c596adf60623/servo/connectors/newrelic.py#L94

servo/connectors/newrelic.py
benburdick: @linkous8 Component in this case would be "canary" - in the servo-harness connector its referenced as `comp_key`

servo/connectors/newrelic.py
blakewatters: I had a long conversation with Ben this afternoon trying to get my head around the specific challenges/requirements for this instance business in the context of Harness.

I have recommended that we explore the idea of using New Relic's support of multiple application names which we can set via the `NEW_RELIC_APP_NAME` environment variable in the form of  `app-name1;app-name2`.

If there are not any blockers that I can't see at the moment (@benburdick is testing these ideas out), I think we can factor out the special snowflakieness of this implementation.

I have also been looking into New Relic labels/tags as another supporting mechanism for some of the stuff we are doing but this probably starts to intersect more with the currently theoretical Harness servox connector

servo/connectors/newrelic.py
blakewatters: We are going to need to revisit the `comp_name`/`component` stuff anyway. The current naming convention precludes us from being able to do multi-connector (i.e. the artist formerly known as aggregation) setups. I think that we may be able to factor out the reliance on the assets endpoint entirely which would be a big win because the data is non-authoritative but this discussion is trending more toward Harness than core New Relic concerns.

I would recommend that to start wading into these waters we simplify the implementation in the core New Relic connector by returning data for all instances unless another connector responds to the `instances` event, getting the rest of the implementation of this port landed, and then bringing up a Harness connector on a new branch that starts with nothing but an `instances` event handler.

We can start with the existing implementation using the IP address hostname computation stuff and then evolve it once we validate some of the changes @benburdick and I have been discussing

servo/connectors/newrelic.py
blakewatters: We should evict this out into a Harness specific connector. We don't want our New Relic connector carrying around implementation details about taking private IP addresses from EC2 instances running in Fargate that have been posted to a private API and are then interpolated into this hostname template which is tightly coupled to internal EC2 instances running in the `us-west-2` data center/availability zone.

We have a couple of distinct but intersecting concerns that are getting gummed up here

blakewatters:
Migrated to standalone repository: https://github.com/opsani/servo-newrelic/pull/4

PR #83, opened 2020-11-12
Feature/wavefront connector

DanielHHowell:
Initial draft of wavefront connector. 

Some notes:

- Added a couple of wavefront formatted types to types.py
- Token is handled as a pydantic.SecretStr, not sure if this is expected/ideal
- Passing pytests, with a small jerry-rig for how the WavefrontConfiguration.generate().yaml() seems to be sorting the items (seen in tests/connectors/wavefront_test.py #L109)

### DIFF
--- a/servo/connectors/wavefront.py
+++ b/servo/connectors/wavefront.py
@@ -27,14 +27,20 @@ class WavefrontMetric(servo.Metric):
     documentation.
     """
 
-    granularity: str = "m"
+    granularity: pydantic.constr(
+        regex=r"^[smhd]"
+    ) = "m"
+
     """The resolution of the query. The granularity resolution determines the number of data points captured across a
     query range. A query's granularity is completely independent from any range durations specified in the WQL expression it evaluates.
     
     Available values: s, m, h, d.
     """
 
-    summarization: str = "LAST"
+    summarization: pydantic.constr(
+        min_length=3,
+        max_length=6
+    ) = "LAST"
     """Summarization strategy to use when bucketing points together.
 
     Available values: MEAN, MEDIAN, MIN, MAX, SUM, COUNT, LAST, FIRST.
@@ -51,17 +57,11 @@ class WavefrontMetric(servo.Metric):
         )
 
 
-# class WavefrontTarget(pydantic.BaseModel):
-#     """WavefrontTarget objects describe targets that are scraped by Wavefront jobs."""
-#
-#     labels: Optional[Dict[str, str]]
-
-
 class WavefrontConfiguration(servo.BaseConfiguration):
     """WavefrontConfiguration objects describe how WavefrontConnector objects
     capture measurements from the Wavefront metrics server.
     """
-    api_key: pydantic.SecretStr
+    api_key: pydantic.SecretStr = '1234567'
     """The API key for accessing the Wavefront metrics API."""
 
     base_url: pydantic.AnyHttpUrl = DEFAULT_BASE_URL
@@ -76,11 +76,6 @@ class WavefrontConfiguration(servo.BaseConfiguration):
 
     Metrics must include a valid query.
     """
-
-    # targets: Optional[List[WavefrontTarget]]
-    # """An optional set of Wavefront target descriptors that are expected to be
-    # scraped by the Wavefront instance being queried.
-    # """
 
     @classmethod
     def generate(cls, **kwargs) -> "WavefrontConfiguration":
@@ -331,10 +326,6 @@ class WavefrontConnector(servo.BaseConnector):
 
         data = response.json()
         self.logger.trace(f"Got response data for metric {metric}: {data}")
-
-        # No status code returned from wavefront response?
-        # if "status" not in data or data["status"] != "success":
-        #     return []
 
         readings = []
 
--- a/servo/connectors/wavefront.py
+++ b/servo/connectors/wavefront.py
@@ -62,7 +62,7 @@ class WavefrontConfiguration(servo.BaseConfiguration):
     """WavefrontConfiguration objects describe how WavefrontConnector objects
     capture measurements from the Wavefront metrics server.
     """
-    api_key: pydantic.SecretStr = '1234567'
+    api_key: pydantic.SecretStr
     """The API key for accessing the Wavefront metrics API."""
 
     base_url: pydantic.AnyHttpUrl = DEFAULT_BASE_URL
--- a/servo/connectors/wavefront.py
+++ b/servo/connectors/wavefront.py
@@ -48,6 +48,7 @@ class WavefrontMetric(servo.Metric):
 
     @property
     def query_escaped(self) -> str:
+        # Not used as such, leaving for now in prom style
         return re.sub(r"\{(.*?)\}", r"{{\1}}", self.query)
 
     def __check__(self) -> servo.Check:
--- a/servo/connectors/wavefront.py
+++ b/servo/connectors/wavefront.py
@@ -61,6 +61,8 @@ class WavefrontConfiguration(servo.BaseConfiguration):
     """WavefrontConfiguration objects describe how WavefrontConnector objects
     capture measurements from the Wavefront metrics server.
     """
+    api_key: pydantic.SecretStr
+    """The API key for accessing the Wavefront metrics API."""
 
     base_url: pydantic.AnyHttpUrl = DEFAULT_BASE_URL
     """The base URL for accessing the Wavefront metrics API.
@@ -147,7 +149,7 @@ class WavefrontRequest(pydantic.BaseModel):
             + f"&e={self.end.timestamp()}"
             + f"&g={self.metric.granularity}"
             + f"&summarization={self.metric.summarization}"
-            + f"&strict=True" # Should probably be a non-configurable property, else query will return points outside the window
+            + f"&strict=True"  # Should probably be a non-configurable property, else query will return points outside the window
         )
 
 
@@ -157,13 +159,6 @@ class WavefrontChecks(servo.BaseChecks):
     """
 
     config: WavefrontConfiguration
-
-    @servo.require('Connect to "{self.config.base_url}"')
-    async def check_base_url(self) -> None:
-        """Checks that the Wavefront base URL is valid and reachable."""
-        async with httpx.AsyncClient(base_url=self.config.api_url) as client:
-            response = await client.get("targets")
-            response.raise_for_status()
 
     @servo.multicheck('Run query "{item.query_escaped}"')
     async def check_queries(self) -> Tuple[Iterable, servo.CheckHandler]:
@@ -181,27 +176,19 @@ class WavefrontChecks(servo.BaseChecks):
             self.logger.trace(
                 f"Querying Wavefront (`{metric.query}`): {wavefront_request.url}"
             )
-            async with httpx.AsyncClient() as client:
-                response = await client.get(wavefront_request.url)
-                response.raise_for_status()
-                result = response.json()
-                return f"returned {len(result)} results"
+
+            async with httpx.AsyncClient(
+                base_url=wavefront_request.url,
+                headers={'Authorization': f'Bearer {self.config.api_key}'},
+            ) as client:
+                try:
+                    response = await client.get()
+                    response.raise_for_status()
+                except (httpx.HTTPError, httpcore._exceptions.ReadTimeout, httpcore._exceptions.ConnectError) as error:
+                    self.logger.trace(f"HTTP error encountered during GET {wavefront_request.url}: {error}")
+                    raise
 
         return self.config.metrics, query_for_metric
-
-    # @servo.check("Active targets")
-    # async def check_targets(self) -> str:
-    #     """Checks that all targets are being scraped by Wavefront and report as
-    #     healthy.
-    #     """
-    #     async with httpx.AsyncClient(base_url=self.config.base_url) as client:
-    #         response = await client.get("/api/v1/targets")
-    #         response.raise_for_status()
-    #         result = response.json()
-    #
-    #     target_count = len(result["data"]["activeTargets"])
-    #     assert target_count > 0, "no targets are being scraped by Wavefront"
-    #     return f"found {target_count} targets"
 
 
 @servo.metadata(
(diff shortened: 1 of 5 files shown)


blakewatters:
Detailed comments forthcoming on the PR itself but a couple of things based on the description:

1. Unless the types that you have added are generally reusable across connectors, move them back into the Wavefront module. We don't anticipate that this connector is going to be super commonly used so I am anticipating that we will spin it out into a standalone repo (which I will help you with, see [servo-webhooks](https://github.com/opsani/servo-webhooks) for an example of what this looks like).
2. Keeping the token in the config in this way is probably fine. The configuration base class dynamically builds environment variable names that can be used to set any field in the configuration. So under something Kubernetes, we can use a Secret that is mounted into the Pod via an environment variable to keep it secure. The `pydantic.SecretStr`  class gives us obfuscation from accidentally leaking it via logging. If a security/risk team pushes back at some point we can look at a generalized secret/credential store outside of the config module but until then I think this setup is sufficient. You can use a `.env` file in your working copy to hold the token and avoid heartburn around accidentally checking it in.
3. Surprised you are seeing weirdness with the config keys but I'll take a look once I get there in the code review.

tests/connectors/wavefront_test.py
```diff
@@ -0,0 +1,459 @@
...
+            assert {
+                "loc": ("metrics",),
+                "msg": "none is not an allowed value",
+                "type": "type_error.none.not_allowed",
+            } in error.errors()
+
+    # Generation
+    def test_generate_default_config(self):
+        config = WavefrontConfiguration.generate()
+        # Ugly patch for assertion until I understand why generate().yaml() metric order is being sorted alphabetically
```
blakewatters, line 111: This is why hate winding up with so many branches in the air but I may have already sorted this out on another branch. I wound up switching to using [ruamel](https://yaml.readthedocs.io/en/latest/) for YAML serialization for some kind of issue with values gettiing mangled but I can't find the bits at the moment.

Ruamel is already a transient dependency through another library on tree, so you may want to try doing a `poetry add ruamel` and then updating `servo.config.AbstractBaseConfiguration.yaml` to pass it through ruamel instead of the stock library.

Ruamel supports round-tripping comments and maintaining ordering and some other stuff so I expect that we'll standardize on it anyway because it would be really nice to emit comments when you generate a config file.

tests/connectors/wavefront_test.py
```diff
@@ -0,0 +1,459 @@
...
+        config = WavefrontConfiguration(
+            base_url="http://localhost:2878", metrics=[metric]
+        )
+        return WavefrontConnector(config=config)
+
+    async def test_describe(self, connector) -> None:
+        described = connector.describe()
+        assert described.metrics == connector.metrics()
+
+    @ respx.mock
```
blakewatters, line 453: little whitespace nit on the decorator

blakewatters (changes requested):
Looks pretty good to me. A couple of minor items in the review feedback that shouldn't take long to mop up.

How was the development experience? Anything we should focus on in the documentation or APIs that gave you trouble?

servo/types.py
```diff
@@ -371,6 +371,14 @@ class Unit(str, enum.Enum):
...
     """
 
+    REQUESTS_PER_MINUTE_WF = "requests/m"
+    """Application throughput in terms of requests processed per minute in Wavefront format.
+    """
+
+    ERRORS_PER_MINUTE_WF = "errors/m"
+    """Application error rate in terms of errors per minute in Wavefront format.
+    """
+
```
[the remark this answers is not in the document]
DanielHHowell: So I am confused how to implement this change. Might need a sample code hint! Units can't be subclassed as an Enum right?

servo/connectors/wavefront.py
```diff
@@ -0,0 +1,352 @@
...
+        return self.metric.granularity
+
+    @property
+    def summarization(self) -> str:
+        return self.metric.summarization
+
+    @property
+    def url(self) -> str:
+        return "".join(
+            self.base_url.rstrip("/")
```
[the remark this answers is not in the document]
DanielHHowell: As far as I can tell it is necessary and probably good to have both representations available?

tests/connectors/wavefront_test.py
DanielHHowell: Ok so I was getting a bit of a headache trying to get ruamel to work. In the meantime I do now have the yaml generating properly with this one-liner:

 ```yaml.add_representer(dict, lambda self, data: yaml.representer.SafeRepresenter.represent_dict(self, data.items()))```

servo/types.py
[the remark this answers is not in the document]
DanielHHowell: There are also some more possible types we need to take in from wavefront once I understand the ideal way to do so

DanielHHowell:
Ok these changes should be in place! I had some confusion with the unit types but consulted with Fred, who informed me they are not actually parsed from the wavefront format and it's up to the user to get it in the right standard format. So I've just replaced the values with standard ones for the default config generation. Let me know if you had something else in mind. 

It's now using `yaml.add_representer(dict, lambda self, data: yaml.representer.SafeRepresenter.represent_dict(self, data.items()))` to correctly parse into yaml. 

SecretStr default value has been removed.

servo/types.py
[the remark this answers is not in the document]
blakewatters: Yeah you can inherit members in a stock Enum -- that's one of the things that the `aenum` library adds (its written by the same dude who did the stdlib enum support).

I will loop back with some reference code after my next meeting

blakewatters:
Sweet! I will do another pass this afternoon

blakewatters:
Apparently we can suppress the default sorting of keys in YAML: https://stackoverflow.com/a/52268128

May just look at doing this globally

DanielHHowell:
> Apparently we can suppress the default sorting of keys in YAML: https://stackoverflow.com/a/52268128
> 
> May just look at doing this globally

Wow can't believe I missed this simple fix. Wheel reinventions 2.0

tests/connectors/wavefront_test.py
blakewatters: You can pull this out. I just added the `sort_keys=False` argument on master (which I was already using in the CLI but missed in the model, derp)

servo/types.py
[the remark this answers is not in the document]
blakewatters: What are the metric types that you want to provide?

We could change the `servo.types.Unit` to having an abstract base class that you could directly inherit from within the Wavefront connector instead of screwing with a new library.

I feel like we have a sorta chicken and egg problem where you don't quite know what I am talking about and I can't quite get your use-case fully in my head.

blakewatters:
> Ok these changes should be in place! I had some confusion with the unit types but consulted with Fred, who informed me they are not actually parsed from the wavefront format and it's up to the user to get it in the right standard format.

I want to push back on this one a bit. One of my goals with ServoX has been to make it as ergonomic and supportive as possible and try to catch as many things as possible at config time rather than having them blow up at runtime.

If there's a format that can be expressed via a regex or whatever that we can check and not rely on the user to format it exactly right, we should go ahead and do that. It's typically just a simple inheritance and a Pydantic validator/constrained type away.

The users configuring this stuff are unlikely to be steeped in Wavefront lore and it's a total bitch when you let something run for 45 minutes and then it blows up for something that could have been validated. I want to eliminate as much of that noise as possible.

DanielHHowell:
> > Ok these changes should be in place! I had some confusion with the unit types but consulted with Fred, who informed me they are not actually parsed from the wavefront format and it's up to the user to get it in the right standard format.
> 
> I want to push back on this one a bit. One of my goals with ServoX has been to make it as ergonomic and supportive as possible and try to catch as many things as possible at config time rather than having them blow up at runtime.
> 
> If there's a format that can be expressed via a regex or whatever that we can check and not rely on the user to format it exactly right, we should go ahead and do that. It's typically just a simple inheritance and a Pydantic validator/constrained type away.
> 
> The users configuring this stuff are unlikely to be steeped in Wavefront lore and it's a total bitch when you let something run for 45 minutes and then it blows up for something that could have been validated. I want to eliminate as much of that noise as possible.

That makes sense. I don't quite have a handle on how to do that with the codebase though, my inheritence attempts were thwarted as the units class flows up to all of the other types classes

servo/types.py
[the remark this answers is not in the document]
DanielHHowell: So the unit types are equivalent, it's just the input strings are different. In sample configs Robert supplied me we havve things like the "request/m" vs. "rpm" and "percent" instead of "%". It's just the string format.

blakewatters:
Roger. I am taking a few mins to spin through Wavefront docs and see if I can get my head around it and advise more intelligently

blakewatters:
Not worth ruminating on -- we have the checks functionality which satisfies my desire to know if we are configured right before running. We can always tune it up further once it's in the field.

There's a couple of broken tests and a conflict with master (guessing from my change to the `sort_keys` arg) and then let's just get it merged.

DanielHHowell:
Ok sounds good. Masters are synced again now. The broken tests are a result of the SecretStr not having a default value. They are all passing with it provided.

blakewatters:
Welp we're not going to merge it with a pile of broken tests to be mopped up immediately.

I think all you need to do is to create a Pytest fixture within your test suite and utilize it instead of directly initializing the configuration class:

```python
@pytext.fixture()
def wavefront_config() -> servo.connectors.wavefront.WavefrontConfiguration:
    return WavefrontConfiguration(base_url="http://localhost:2878", metrics=[], api_key="abcd12345")
```

And then just reference `wavefront_config()` in the following broken cases:

```
FAILED tests/connectors/wavefront_test.py::TestWavefrontConfiguration::test_supports_localhost_url
FAILED tests/connectors/wavefront_test.py::TestWavefrontConfiguration::test_supports_cluster_url
FAILED tests/connectors/wavefront_test.py::TestWavefrontConfiguration::test_api_url
FAILED tests/connectors/wavefront_test.py::TestWavefrontConfiguration::test_generate_default_config
FAILED tests/connectors/wavefront_test.py::TestWavefrontConfiguration::test_base_url_is_rstripped
FAILED tests/cli_test.py::test_init_from_scratch - pydantic.error_wrappers.Va...
ERROR tests/connectors/wavefront_test.py::TestWavefrontChecks::test_check_queries
ERROR tests/connectors/wavefront_test.py::TestWavefrontConnector::test_measure
ERROR tests/connectors/wavefront_test.py::TestWavefrontConnector::test_describe
```

blakewatters:
I suspect that `tests/cli_test.py::test_init_from_scratch` is breaking because your generate() implementation returns an incomplete configuration.

It is reasonable to supply a sentinel value for `api_key` in the `generate` method so that it gets written out to the generated config file (something like `replace-me-with-your-key` or whatever)

DanielHHowell:
Ahh ok, sorry for the confusion there. Thanks for the pointers + patience, those changes are now in place

servo/connectors/wavefront.py
```diff
@@ -0,0 +1,353 @@
...
+    """
+
+    summarization: pydantic.constr(
+        min_length=3,
+        max_length=6
+    ) = "LAST"
+    """Summarization strategy to use when bucketing points together.
+
+    Available values: MEAN, MEDIAN, MIN, MAX, SUM, COUNT, LAST, FIRST.
+    """
```
blakewatters, line 47: We should turn this one into an enum instead of doing the pseudo-check with line lengths. Pydantic will constrain the values to members of the enumeration and if you also inherit from `str`, they can be interpolated directly into URL query strings, log messages, etc:

```python
class Summarization(str, enum.Enum):
    mean = "MEAN"
    median = "MEDIAN"
    min = "MIN"
    max = "MAX"
    sum = "SUM"
    count = "COUNT"
    last = "LAST"
    first = "FIRST"
```

servo/connectors/wavefront.py
```diff
@@ -0,0 +1,353 @@
...
+
+    granularity: pydantic.constr(
+        regex=r"^[smhd]"
+    ) = "m"
+
+    """The resolution of the query. The granularity resolution determines the number of data points captured across a
+    query range. A query's granularity is completely independent from any range durations specified in the WQL expression it evaluates.
+    
+    Available values: s, m, h, d.
+    """
```
blakewatters, line 38: I'd probably argue for turning this into an enum as discussed below that Pydantic will constrain it for us and its self-documenting what the available values are.

servo/connectors/wavefront.py
```diff
@@ -0,0 +1,353 @@
...
+                    granularity="m",
+                    summarization="LAST",
+                ),
+                WavefrontMetric(
+                    "error_rate",
+                    servo.Unit.COUNT,
+                    query="avg(ts(appdynamics.apm.transactions.errors_per_min, env=foo and app=my-app))",
+                    granularity="m",
+                    summarization="LAST",
+                ),
```
blakewatters, line 106: With the enum changes recommended above, I think this reads much more expressively:

```python
WavefrontMetric(
    "throughput",
    servo.Unit.REQUESTS_PER_MINUTE,
    query="avg(ts(appdynamics.apm.overall.calls_per_min, env=foo and app=my-app))",
    granularity=Granularity.minute,
    summarized_by=Summarization.last,
),
WavefrontMetric(
    "error_rate",
    servo.Unit.COUNT,
    query="avg(ts(appdynamics.apm.transactions.errors_per_min, env=foo and app=my-app))",
    granularity=Granularity.minute,
    summarized_by=Summarization.last,
),
```

servo/connectors/wavefront.py
```diff
@@ -0,0 +1,353 @@
...
+                except (httpx.HTTPError, httpcore._exceptions.ReadTimeout, httpcore._exceptions.ConnectError) as error:
+                    self.logger.trace(f"HTTP error encountered during GET {wavefront_request.url}: {error}")
+                    raise
+
+        return self.config.metrics, query_for_metric
+
+
+@servo.metadata(
+    description="Wavefront Connector for Opsani",
+    version="0.0.1",
```
blakewatters, line 194: You can probably jump this to 1.0.0 🎉

blakewatters (approved):
Added a few final comments for fit and finish

DanielHHowell:
Excellent! That does constrain better and look a lot cleaner

blakewatters:
Just sent it down to master. Welcome aboard!

PR #85, opened 2020-11-22
Enhanced error module and support for rejections

blakewatters:


blakewatters:
Digging into the backend code, it looks like most of the stuff specified in the spec is not implemented or validated. I am going to cut back on the error classes and not worry about covering the spec.

blakewatters:
Merging down. Solid test coverage really needs bits from another branch

PR #106, opened 2020-12-09
Feature/environment mode

linkous8:
Logic to manage annotations of kubernetes deployments to facilitate requesting of environment change as well as notifying OKO


linkous8:
Efforts moved to https://github.com/opsani/servox/tree/feature/environment-mode-v2

Issue #80, opened 2020-11-11
DeploymentOptimization.apply fails due to error in kubernetes_asyncio library


linkous8:
I've reproduced the issue on master for thoroughness but the log ([master_output.txt](https://github.com/opsani/servox/files/5525750/master_output.txt)) isn't very useful due to https://github.com/opsani/servox/issues/79

See this log from my dev branch for more complete error information: [dev_branch_output.txt](https://github.com/opsani/servox/files/5525763/dev_branch_output.txt)

linkous8:
Running `poetry lock` to update dependencies seems to have rectified the error kubernetes_asyncio was giving. However it now times out with no explicit error given: [new_lock_output.txt](https://github.com/opsani/servox/files/5526067/new_lock_output.txt)

```
2020-11-11 12:34:16.206 | INFO     | servo.connectors.kubernetes:apply:2278 | kubernetes[adjust] - Applying 3 Kubernetes adjustments: [web-main/main.cpu=0.5, web-main/main.mem=0.5, web-main/main.replicas=1]
2020-11-11 12:34:16.209 | INFO     | servo.connectors.kubernetes:apply:2285 | kubernetes[adjust] - adjusting web-main/main: web-main/main.cpu=0.5
2020-11-11 12:34:16.210 | INFO     | servo.connectors.kubernetes:apply:2285 | kubernetes[adjust] - adjusting web-main/main: web-main/main.mem=0.5
2020-11-11 12:34:16.211 | INFO     | servo.connectors.kubernetes:apply:2285 | kubernetes[adjust] - adjusting web-main/main: web-main/main.replicas=1
2020-11-11 12:34:16.211 | DEBUG    | servo.connectors.kubernetes:apply:2294 | kubernetes[adjust] - waiting for adjustments to take effect on 1 optimizations
2020-11-11 12:34:16.231 | INFO     | servo.connectors.kubernetes:apply:1865 | kubernetes[adjust] - Using label_selector=app=web,role=main, resource_version=2335
2020-11-11 12:39:16.311 | ERROR    | servo.connectors.kubernetes:apply:2316 | kubernetes[adjust] - timed out after 5m waiting for adjustments to apply
2020-11-11 12:39:16.312 | INFO     | servo.connectors.kubernetes:rollback:1761 | kubernetes[adjust] - adjustment failed: rolling back deployment... ()
[...]

PR #107, opened 2020-12-11
Implemented is_ready on optimization classes and rejection checks during waits

linkous8:
NOTE: this hasn't been completely tested (testing is partially blocked by #80) but I wanted to get the ball rolling on discussion and review anyway

@blakewatters I was unable to find the implementation for the `is_ready` method on the `KubernetesOptimizations` class so I provided one and added some extra rejection checks to bring the logic to feature parity with the legacy implementation. LMK your thoughts and any requests for changes


blakewatters:
```
@Blake thanks for merging part of the rejection implementation into the main branch!
However, there were some items from my branch that I was wondering if we could incorporate as well. Let me know if I may be of assistance in doing so:
I wrote several pytests that will trigger rejections and made some fixes to errors.py so that the tests dont fail due to missing context variables
 The configured timeout is not being set on the DeploymentOptimization and CanaryOptimization apply methods so only the KubernetesOptimization's wait_for is actually capable of timing out (Further , KubernetesOptimization's timeout should be extended so that it will not be the first timeout triggered)
Finally, I've noticed the servo errors don't seem to show in the logs of the WebUI like they used to. It seems handle_error is not communicating them to the backed (important for rejection status) and so I had some changes to raise errors after handling them but LMK if I have the wrong idea here
Thanks again! :slightly_smiling_face:
```

linkous8:
One more essential item; [this watch](https://github.com/opsani/servox/blob/65d5c7624348347f2537bd33b0662177126da4f2/servo/connectors/kubernetes.py#L1647) doesn't raise a TimeoutError so we need to raise an error if the timeout elapses and instead of returning cleanly

I listed what I thought were essential but the PR also includes pod scheduling failure checks that can be omitted if need be since they will be rejected by the timeout logic anyway

servo/connectors/kubernetes.py
```diff
@@ -1446,6 +1449,37 @@ async def get_pods(self) -> List[Pod]:
...
+            label_selector = self.obj.spec.selector.match_labels
+            rs_list:kubernetes_asyncio.client.V1ReplicasetList = await api_client.list_namespaced_replica_set(
+                namespace=self.namespace, label_selector=selector_string(label_selector), resource_version=self.resource_version
+            )
+
+        # Verify all returned RS have this deployment as an owner
+        rs_list = [
+            rs for rs in rs_list.items if rs.metadata.owner_references and any(
+                ownRef.kind == "Deployment" and ownRef.uid == self.obj.metadata.uid
+                for ownRef in rs.metadata.owner_references)]
```
blakewatters, line 1469: This one is pretty hard to read. Non-blocking for merge but there's prolly a tighter cut

servo/connectors/kubernetes.py
```diff
@@ -1446,6 +1449,37 @@ async def get_pods(self) -> List[Pod]:
...
+        if not rs_list:
+            raise servo.ConnectorError('Unable to locate replicaset(s) for deployment "{self.name}"')
+        latest_rs = sorted(rs_list, key= lambda rs: rs.metadata.resource_version, reverse=True)[0]
+
+        return [
+            pod for pod in await self.get_pods()
+            if any(
+                ownRef.kind == "ReplicaSet" and ownRef.uid == latest_rs.metadata.uid
+                for ownRef in pod.obj.metadata.owner_references
+            )]
```
blakewatters, line 1479: Also not the clearest read to my eye. Should this `any` check here and above be encapsulated into a method somewhere?

It's hard to tell what its actually trying to express even beyond the density of nested list comprehensions

servo/connectors/kubernetes.py
```diff
@@ -1746,6 +1787,22 @@ def _check_conditions(self, conditions: List[kubernetes_asyncio.client.V1Deploym
                         f"unknown deployment status condition: {condition.status}"
                     )
 
+    async def _check_pod_conditions(self):
+        pods = await self.get_latest_pods()
+        unschedulable_pods = [
+            pod for pod in pods
+            if pod.obj.status.conditions and any(
+                cond.reason == "Unschedulable" for cond in pod.obj.status.conditions
+            )]
```
blakewatters, line 1796: Maybe tuck this into an `get_unschedulable_pods()` method?

servo/connectors/kubernetes.py
```diff
@@ -1746,6 +1787,22 @@ def _check_conditions(self, conditions: List[kubernetes_asyncio.client.V1Deploym
...
+        if unschedulable_pods:
+            pod_fmts = [] # [f"{pod.obj.metadata.name} - {', '.join(cond.message for cond)}" for pod in unschedulable_pods]
```
blakewatters, line 1798: What's up with the comment?

servo/connectors/kubernetes.py
```diff
@@ -2087,9 +2144,8 @@ async def handle_error(self, error: Exception, mode: "FailureMode") -> bool:
                 must filter failure modes before calling the superclass implementation.
         """
         if mode == FailureMode.crash:
-            raise RuntimeError(
-                "an unrecoverable failure occurred while interacting with Kubernetes"
-            ) from error
+            self.logger.error(f"an unrecoverable failure occurred while interacting with Kubernetes: {error.__class__.__name__} - {str(error)}")
+            raise error
```
blakewatters, line 2148: `RuntimeError` was kind of a lazy solution here but my intent was to use exception chaining so that you would have a known surface area to trap with an except block since its hard to know what's gonna bubble up here

servo/connectors/kubernetes.py
```diff
@@ -2100,19 +2156,21 @@ async def handle_error(self, error: Exception, mode: "FailureMode") -> bool:
...
                 await self.destroy(error)
-                return True
+
+            else:
+                # Trap any new modes that need to be handled
+                raise NotImplementedError(
+                    f"missing error handler for failure mode '{mode}'"
+                ) from error
+
+            raise error # Always communicate errors to backend unless ignored
```
blakewatters, line 2169: Isn't this code unreachable?

servo/connectors/kubernetes.py
```diff
@@ -2710,7 +2769,7 @@ async def apply(self, adjustments: List[servo.Adjustment]) -> None:
                         *list(map(lambda a: a.apply(), self.optimizations)),
                         return_exceptions=True,
                     ),
-                    timeout=timeout.total_seconds(),
+                    timeout=timeout.total_seconds() + 60, # allow sub-optimization timeouts to expire first
```
blakewatters, line 2772: Prolly want to put the 60 into a constant since its referenced multiple times as a literal

blakewatters (approved):

servo/connectors/kubernetes.py
linkous8: Sounds reasonable, we'd just need to add the logic to unwrap chained exceptions since, currently, the top-most exception in the chain is the only one that I saw show up on the Web UI

servo/connectors/kubernetes.py
linkous8: Its hit when FailureMode is rollback or destroy

servo/connectors/kubernetes.py
linkous8: Agree

PR #116, opened 2020-12-28
Refreshed units per feedback. 

lswank:
ms is now milliseconds. 
Count, the unitless unit, is now no longer a unit.
rps is now req / sec (although we could make it more explicit).

This is another smaller pull request mostly to see if I can get tests to pass nicely. The only other place where I explicitly did not change "ms" related code was in the utils library code.


servo/types.py
```diff
@@ -529,10 +529,10 @@ class Unit(str, enum.Enum):
     count = ""
     rate = ""
     percentage = "%"
-    milliseconds = "ms"
+    milliseconds = "milliseconds"
     bytes = "bytes"
     requests_per_minute = "rpm"
-    requests_per_second = "rps"
+    requests_per_second = "req / sec"
```
blakewatters, line 535: I feel like we kinda have to define a standard here -- having `rpm` and `req / sec` feels dissonant.

I'm also a little iffy on if we want to introduce unit abbreviations that are this long. We're gonna append it on to log messages and string representations of the metrics for output and it's a lot of horizontal real estate for something that is pretty well understood at a glance.

tests/types_test.py
```diff
@@ -119,28 +119,28 @@ def test_parse_measure_command_response_including_units() -> None:
                     "unit": "%",
                 },
                 "latency_total": {
-                    "unit": "ms",
+                    "unit": "milliseconds",
```
blakewatters, line 122: This one is weird to me too when you consider how thoroughly we have wired against Golang style `Duration` strings all over the codebase and into the config.

If we need/want expanded long-form units we can probably do something clever and sexy with an enum subclass

lswank: I actually don't buy the notion that people won't interpret "ms" at a glance properly.￼ It's pretty standard across Posix-everything.

PR #129, opened 2021-01-13
Introduce support for pub/sub messaging

blakewatters:
This is a snapshot of the work-in-progress pub/sub functionality designed to support **intermittent traffic** and **OLAS** integrations with ServoX.

The module provides a lightweight publisher/subscriber abstraction that facilitates collaboration between connectors (or external systems) based on the exchange of messages. Publishers and subscribers are decoupled through participation in an exchange, which acts as a message broker. Channels have simple textual names (e.g., `metrics` or `metrics.http.latencies`) and subscribers can utilize string glob patterns (`metrics.*`) or regular expression (`^metrics\.(http|mysql).*$`) selectors to target the messages that they are interested in.

The implementation is decoupled from the connector and event core modules to enable interoperability without having to bend integration targets into the connector interface if it doesn't make sense. The `Message` interface has also been designed to require serialization as text, JSON, or YAML representations to facilitate usage across network and process boundaries (it is hard to retrofit these properties so better to have them now).

To make the common use-cases delightful and easy, there is a mixin provided that enables an even higher level set of abstractions that will be mixed in by default to the `servo.BaseConnector` base class.

## Quick Example

This is an example of the fully integrated solution embedded within connector instances that become available out of the box.

```python
class PrometheusConnector(servo.BaseConnector):
    async def startup(self) -> None:
        @self.publisher("metrics.prometheus", every="15s")
        async def _publish_metrics(publisher: servo.pubsub.Publisher) -> None:
            metrics = await self.query_prometheus_for_metrics()
            servo.logger.info(f"Publishing {len(metrics)} metrics to Subscribers...")
            publisher(servo.pubsub.Message(json=metrics))


class OLASConnector(servo.BaseConnector):
    async def startup(self) -> None:
        @self.subscriber("metrics.prometheus.*")
        async def _process_metrics(message: servo.pubsub.Message, channel: servo.pubsub.Channel) -> None:
            await self.olas.do_something_with_metrics(message.json())
```

The module is pretty well documented at this point so look at the PR for the underlying details. The tests and remaining functionality coming in shortly.


servo/pub_sub.py
```diff
@@ -0,0 +1,430 @@
...
+        self._subscribers.clear()
+
+    async def _process_queue(self) -> None:
+        while True:
+            message, channel = await self._queue.get()
+            if message is None:
+                # Exit condition
+                break
+
+            results = asyncio.gather(
```
opcli, line 305: I assume you run one exchange for all the servox message queue. Otherwise if one exchange per channel, there is no need to use the channel any more.

Here it has the issue of serializing all the message and caller. For example, let's say there are two channel of message is totally unrelated. Let's suppose channel "a" has message "A" then channel "b" has message "B". 

The processing will force message "B" does not start to get process before *all* of the call back of the message "A" completes. The is not optimal because "b" channel might have nothing to do with "a". That means any process handler can stall the processing pipe which can be a design draw back.

opcli:
I have some feedback for the given example code:
```

class PrometheusConnector(servo.BaseConnector):
    async def startup(self) -> None:
        @self.publisher("metrics.prometheus", every="15s")
        async def _publish_metrics(publisher: servo.pubsub.Publisher) -> None:
            metrics = await self.query_prometheus_for_metrics()
            servo.logger.info(f"Publishing {len(metrics)} metrics to Subscribers...")
            publisher(servo.pubsub.Message(json=metrics))
```
Some feedback for the publisher subscribe API:

1. Don't have to use the "decorator" input argument to access the publisher API

  I wish there is a simpler way to access a to the publisher. Just like an module function call.
  Why force to hide the publisher in call back function arguments?

2. Allow publisher run into their own async task.
   This is relate to the API design as well. Currently he publisher call back have to be run in the publisher's
   API call back, in the publisher's module's task.

   I wish the publisher API allows the caller to live in their own task. This will allow much more flexible
   when a metrics should be published.

   For example, Olas does not use a fix internal on the sleep internal. The interval is calculated base on
   the metrics it collects.

   Allowing the publisher run on their own async task, the Olas can use "async.sleep()" to determine how
   long the interval is.
   That will result in a simpler and more flexible API design.

   It will make the code easier to debug as well, as Olas publisher task will show up in the call stack if some thing
   wrong happen.

As for the subscriber example
```
class OLASConnector(servo.BaseConnector):
    async def startup(self) -> None:
        @self.subscriber("metrics.prometheus.*")
        async def _process_metrics(message: servo.pubsub.Message, channel: servo.pubsub.Channel) -> None:
            await self.olas.do_something_with_metrics(message.json())
```
3. I have the same wish allowing the subscriber to run in the context of their own task as well.
    The "asyncio.Queue" can already does that. e.g. Olas async use that now.

   If the subscriber can run on its own async task. Not only does it make the processing of the message easier to debug.
   It also address the issue I raise in the inline commenting about "channel a message can stall channel b message". 

I also wish, if possible, there is less boiler plat code to make it work. But I also notice that is my personal preference.

servo/pub_sub.py
blakewatters: There will be one `Exchange` shared across all Connectors via the mixin -- just like the existing event bus.

You are correct that there is a serialization of message processing here which is really a convenience for handling the exceptions. There are other approaches that can yield more concurrency with a bit more leg-work. Noted for review

blakewatters:
Hi Chris -

Trying to quote here but it may or may not come out very readable :-)

>     1. Don't have to use the "decorator" input argument to access the publisher API
> 
> 
> I wish there is a simpler way to access a to the publisher. Just like an module function call.
> Why force to hide the publisher in call back function arguments?

You have full access to the underlying API by importing the module. Via the `self.exchange` property, you can directly call the `create_publisher`, `create_subscriber`, `create_channel`, etc underlying APIs and more directly orchestrate things.

The decorator hides the boilerplate and optimizes for the common cases that I see: publishing events on a fixed interval and consuming them via subscription whenever they happen. The decorators let you take a normal function and transform them into a publisher/subscriber without needing to worry about the underlying details and intermediary objects.

I don't see what is not simple about this -- do you have some proposed alternative APIs in mind?

> 
>     1. Allow publisher run into their own async task.
>        This is relate to the API design as well. Currently he publisher call back have to be run in the publisher's
>        API call back, in the publisher's module's task.
>        I wish the publisher API allows the caller to live in their own task. This will allow much more flexible
>        when a metrics should be published.
>        For example, Olas does not use a fix internal on the sleep internal. The interval is calculated base on
>        the metrics it collects.
>        Allowing the publisher run on their own async task, the Olas can use "async.sleep()" to determine how
>        long the interval is.
>        That will result in a simpler and more flexible API design.
>        It will make the code easier to debug as well, as Olas publisher task will show up in the call stack if some thing
>        wrong happen.

There's nothing here that precludes you from doing what you describe. You can directly create a `Publisher` object via the `Exchange` interface and precisely control how and when the publication is executed. The decorator just embraces the pattern of publishing on a fixed interval because I think that's the 80% common case. Prometheus aggregates metrics from its targets on a defined scrape interval so pinging it off-cycle from its scrape interval isn't going to provide any new or different data.

I can easily make the `publisher` decorator allow you to omit the `every` parameter and take direct control over the sleeping if non-fixed intervals actually have a utility that I am not yet seeing.

I don't understand the feedback about the async task or call-stacks. If you call into your own OLAS-specific code from within a publisher or subscriber and it crashes you will see it in the call stack. I am very intentionally trying not to expose the details of the underlying async tasks and queues because they are just implementation details.

> 
> 
> As for the subscriber example
> 
> ```
> class OLASConnector(servo.BaseConnector):
>     async def startup(self) -> None:
>         @self.subscriber("metrics.prometheus.*")
>         async def _process_metrics(message: servo.pubsub.Message, channel: servo.pubsub.Channel) -> None:
>             await self.olas.do_something_with_metrics(message.json())
> ```
> 
>     1. I have the same wish allowing the subscriber to run in the context of their own task as well.
>        The "asyncio.Queue" can already does that. e.g. Olas async use that now.
>        If the subscriber can run on its own async task. Not only does it make the processing of the message easier to debug.
>        It also address the issue I raise in the inline commenting about "channel a message can stall channel b message".
> 
> 
> I also wish, if possible, there is less boiler plat code to make it work. But I also notice that is my personal preference.

Again I don't understand the feedback about the async task context, please elaborate. The stalled processing is a pretty minor issue that I will address in the next revision -- it really only matters if you are going to consume a subscribed message and immediately do another potentially long-running async task blindly in the subscriber processing block.

I also do not understand the comment about the boilerplate code. The decorator based example code that I shared has like zero fat on it and no underlying implementation details are exposed. Connectors that want to publish or subscribe to messages can just write a simple function, decorate it, and they are all wired in together. Because the `Exchange` is rooted on the shared connector hierarchy like the event bus, it is automatically multi-servo aware so there can be an arbitrary number of exchanges running and you never need to even think about nor care about it.

blakewatters:
@opcli the core implementation is now complete and the tests are pushed to the branch. They should clear up any ambiguity about how the APIs fit together.

The mixin just needs to be tested and integrated onto the `BaseConnector` class and I think this thing is ready for action.

opcli:
Hi Blake,

Please fully understanding my comments.

> You have full access to the underlying API by importing the module. Via the `self.exchange` property, you can directly call the `create_publisher`, `create_subscriber`, `create_channel`, etc underlying APIs and more directly orchestrate things.
> 
> The decorator hides the boilerplate and optimizes for the common cases that I see: publishing events on a fixed interval and consuming them via subscription whenever they happen. The decorators let you take a normal function and transform them into a publisher/subscriber without needing to worry about the underlying details and intermediary objects.
> 
> I don't see what is not simple about this -- do you have some proposed alternative APIs in mind?
Yes, I do. It is already in the Olas branch.

Please see the olas event queue to understand how the current Olas event streaming works.

https://github.com/opsani/olas-prototype/blob/master/frontend/olas/eventqueue.py

It is just a very thin layer on top of the asyncio.Queue().

```
class EventQueue:
    def __init__(self):
        self.queues = {}
        # queues are a map from event type to a queue list.

    def put_event(self, event_type, event):
        for queue in self.queues.get(event_type, []):
            queue.put_nowait((event_type, event))

    def subscribe_event(self, event_type):
        queue = asyncio.Queue()
        self.queues.setdefault(event_type, []).append(queue)
        return queue

```

That is it. Very simple and effective.

> There's nothing here that precludes you from doing what you describe. You can directly create a `Publisher` object via the `Exchange` interface and precisely control how and when the publication is executed. The decorator just embraces the pattern of publishing on a fixed interval because I think that's the 80% common case. Prometheus aggregates metrics from its targets on a defined scrape interval so pinging it off-cycle from its scrape interval isn't going to provide any new or different data.
> 
> I can easily make the `publisher` decorator allow you to omit the `every` parameter and take direct control over the sleeping if non-fixed intervals actually have a utility that I am not yet seeing.
> 
> I don't understand the feedback about the async task or call-stacks. If you call into your own OLAS-specific code from within a publisher or subscriber and it crashes you will see it in the call stack. I am very intentionally trying not to expose the details of the underlying async tasks and queues because they are just implementation details.

You miss the point that the current example does not do what Olas want to do. Which is having a streaming
event and allow Olas control the interval of each event. Please update the example how to do that in your
example code.

For the refer to the Olas example, here is how it is done:
https://github.com/opsani/olas-prototype/blob/master/frontend/olas/controller.py#L856
```

async def control_task(self):
       ...
        while 1:
            await self.sleep_loop_interval()
            ...
            pods = await self.get_pods()
            metrics = await self.collect_metrics(pods)
            eventqueue.eventQueue.put_event("OlasMetrics", metrics)

```
In this way, Olas can arrange different interval of events.

Notice the difference here. In your review. Publisher owns the task that call the publisher call back.
In Olas example, Olas owns task that perform the event publishing. That makes the Olas have a 
more coherent task context of how it publish the events.

Can you provide an example how to do the equivalent in Publisher/Exchange API?

> > I also wish, if possible, there is less boiler plat code to make it work. But I also notice that is my personal preference.
> 
> Again I don't understand the feedback about the async task context, please elaborate. The stalled processing is a pretty minor issue that I will address in the next revision -- it really only matters if you are going to consume a subscribed message and immediately do another potentially long-running async task blindly in the subscriber processing block.

Let me provide you the example of how current Olas does event processing.
Please notice the difference compare to the review approach regarding who owns the task that
processing the event.

In the Olas example, Olas owns the processing task. It is simple to understand.
In your review commit, the exchange owns the processing task.

```
    async def process_event_task(self):
        queue = self.eventqueue
        while 1:
            event_type, metrics  = await queue.get()
            logging.info("Process event {metrics}")
            await self.control_with_metrics(metrics)

```

> 
> I also do not understand the comment about the boilerplate code. The decorator based example code that I shared has like zero fat on it and no underlying implementation details are exposed. Connectors that want to publish or subscribe to messages can just write a simple function, decorate it, and they are all wired in together. Because the `Exchange` is rooted on the shared connector hierarchy like the event bus, it is automatically multi-servo aware so there can be an arbitrary number of exchanges running and you never need to even think about nor care about it.

So does the Olas version of event queue. I see the biggest issue is that, servox using decorator for EVERY infrastructure
related API. That cause a lot of boilerplate code. Please consider example like asyncio module. Which is provide by python.
"asyncio" does not using much decorator at all, but still enable the user to do very flexible event loops.

There is harm by over using the decorator for every thing. It actually produces less readable code.
If you are interested, I can show you how to reduce the boilerplate code, just like the event queue example I give.
I can also zoom if that helps.

blakewatters:
Hi Chris -

Couple of things...

1. You may need to slow down a bit and actually read through the code and test cases. Every single piece of feedback that you have presented has been methodically addressed and integrated into the design. There are several different API affordances and levels of abstraction that allow you to trade off convenience for control.

I updated the decorator interface to support execution without timed repetitions based on your feedback. You can omit the `every` parameter and take control over the sleep intervals. You can skip the decorators and create your own Publisher/Subscriber objects. If you really want to just deal with the lowest level details possible, you can just attach a listener to the queue. 

I have thought about and tried to balance all of these concerns. Code is at the end of the post.

2. You should be thinking more broadly. My only mission in life right now is not to enable OLAS and give you what you want. We are developing several other overlapping components, accelerating the POV/sales cycle, and I am trying to get a number of engineers up to speed and enable them with robust, well documented, extensively tested functionality that lets them focus on delivering value at a high spin.

I have been working diligently to integrate your needs alongside all the other concerns on my plate, our ambitions for the future, and the reality of trying to scale up a team in the face of a fair bit of essential complexity that needs to be thoughtfully managed.

It seems like you are coming with an attitude that there is a 20 line module global solution to these engineering challenges that I am either unable or unwilling to contemplate. If so, you are incorrect.

3. The assertion that ServoX is overly using decorators and harming code readability just does not resonate with me. The whole point of decorators is to enhance readability and semantics by exposing the stuff that matters and abstracting away what doesn't. What code is giving you this much heartache?

Much like `asyncio`, I have been assembling things along a set of lower-level and higher-level APIs. The design ethos is to make the easy things trivial, challenging things easy, hard things accessible, and crazy shit possible if you really want to take your life into your own hands. Every decorator and higher-level API is composed of well designed underlying APIs that are also at your fingertips at every moment.

4. Your event loop code is just not unusable in ServoX. You have completely ignored life-cycle management, collaboration/coordination between components, testability, error handling, documentation, yada yada yada. Pretending like a bunch of stuff that you would rather not care about does not make it irrelevant to the rest of us. There is a difference between simple and primitive.

----

Alright, let's answer your sample code queries. You can subscribe and publish messages in a couple of ways:

```python
class ExampleConnector(servo.BaseConnector):
    async def startup(self) -> None:
        ## Publisher examples

        # Manually sleep the loop
        @self.publish('metrics.olas')
        async def _publish_olas_metrics() -> None:
            pods = await self.get_pods()
            metrics = await self.collect_metrics(pods)
            self.olas.put_event("OlasMetrics", metrics)

            sleep_duration = await self.olas.sleep_loop_interval()
            await asyncio.sleep(sleep_duration)

        # Schedule the loop
        @self.publish('metrics.olas', every='15s')
        async def _publish_olas_metrics_scheduled() -> None:
            pods = await self.get_pods()
            metrics = await self.collect_metrics(pods)
            self.olas.put_event("OlasMetrics", metrics)

        # Cancel it by name (defaults to the func name)
        self.cancel_publishers('_publish_olas_metrics_scheduled')

        # or cancel them all
        self.cancel_publishers()

        # Publish within an execution block
        async with self.publish('metrics.olas') as publisher:
            for _ in range(5):
                pods = await self.get_pods()
                metrics = await self.collect_metrics(pods)
                await publisher(metrics)
                self.olas.put_event("OlasMetrics", metrics)

        # Don't even use a loop or a block at all 
        # -- just give me an object and I'll publish when I want
        publisher = self.publish('metrics.olas')
        await publisher(servo.pubsub.Message(text="What's up?"))

        # We're done with it now
        publisher.cancel()

        ##
        # Subscriber examples
        
        @self.subscribe('metrics.prometheus')
        async def _process_message(message: servo.pubsub.Message, channel: servo.pubsub.Channel) -> None:
            await self.olas.do_something(message.json())
            sleep_duration = await self.olas.sleep_loop_interval()
            await asyncio.sleep(sleep_duration)

        # Or you could do it like this
        async with self.subscribe('metrics.prometheus') as subscription:
            async for message, channel in subscription:
                print(f"New Message in #{channel.name}: {message}")

        ##
        # You hate decorators and abstractions? Let's do it the hard way.

        exchange = servo.pubsub.Exchange()
        publisher = exchange.create_publisher('metrics.production.http')
        subscriber = exchange.create_subscriber('metrics.*.http*')
        event = asyncio.Event()

        async def _publisher_func() -> None:
            # Wait for subscriber registration
            await event.wait()

            for i in range(10):
                message = servo.pubsub.Message(text=f'Test Message #{i}')
                await publisher(message)

        async def _subscriber_func() -> List[servo.pubsub.Message]:
            async with exchange.subscribe('metrics*') as subscription:
                # Trigger the Publisher to begin sending messages
                event.set()

                async for message, channel in subscription:
                    messages.append(message)

                    if len(messages) == 10:
                        subscription.stop()

            return messages

        results = await asyncio.wait_for(
            asyncio.gather(_publisher_func(), _subscriber_func()),
            timeout=3.0
        )
            
```

Less academically, you can take a look at this changeset which made the Vegeta Connector start broadcasting load generation metrics whenever it is running. 

It's like 10 lines of code because of the investment in the underlying infrastructure.  See https://github.com/opsani/servox/blob/c0590b83fc7ca272d05f4f620f424e886a0abf74/servo/connectors/vegeta.py#L342-L346 and https://github.com/opsani/servox/blob/c0590b83fc7ca272d05f4f620f424e886a0abf74/tests/integration/pubsub_test.py#L37-L50

All the lame and tough stuff like setup, teardown, exception handling, etc. are handled in the right way, making fairly advanced programming paradigms much more accessible.

There's a ton of shit in here that I have wired up to make it easy, fast, fun, and safe to develop against.

Cheers,
Blake.

opcli:
Hi Blake,

Sorry for the late reply. Have been busy with other stuff. I will only comment on the area
I think is still problematic. I actually don't have time to go through the rest of your code
yet. This code has grown significantly. 

Please look at here:

```

    async def _process_queue(self) -> None:
        while True:
            message, channel = await self._queue.get()
            if message is None:
                # Exit condition
                break

            # Notify subscribers in a new task to avoid blocking the queue
            asyncio.create_task(_deliver_message_to_subscribers(message, channel, self._subscribers))

            self._queue.task_done()

```

It is around the same area of code where I found the first issue that message A will halt message B behavior.
I still question the design choice of letting the exchange own the task rather than subscriber own the task
like the "async.Queue()" does.

Here is how the bug can be triggered. Let's say producer a produce 3 message A1 A2 A3 at the same time
before the exchange have any chance to run. The subscriber of this message on the channel should get A1
A2 A3, in that order.

Here is the bug, because you create a new task for EACH message. There is actually no  guarantee the
subscriber's callback will execute the in the order of A1 A2 A3. Because all 3 task are available. It is up to
the async scheduler to schedule the 3 task at its choice. If A1 gets block, it is possible A3 get execute and
complete before A2.

As soon as you schedule 3 different task, you lose the ordering of the task. That is why for each subscriber,
a separate async.Queue() is needed. Each subscriber will have its own task, not each message. Since each
subscriber need to have its own task, you might just let subscriber have that task.

BTW, this kind of bug is very subtle and very hard to debug if it actually triggered. Most of the
time the system runs fine, only when it is buys and every once a while it happen, and hard to reproduce.

I still think expose the separate "async.Queue" to the subscriber is much simpler and less boiler plate
code to expose a call back handler.

Make the code simple and easy to understand is a good thing, it helps your quest.
I don't see the conflict on that.

blakewatters:
> Hi Blake, > > Sorry for the late reply. Have been busy with other stuff. I
will only comment on the area > I think is still problematic. I actually don't
have time to go through the rest of your code > yet. This code has grown
significantly. > > Please look at here: > > ``` > >     async def
_process_queue(self) -> None: >         while True: >             message,
channel = await self._queue.get() >             if message is None: > # Exit
condition >                 break > >             # Notify subscribers in a new
task to avoid blocking the queue >
asyncio.create_task(_deliver_message_to_subscribers(message, channel,
self._subscribers)) > >             self._queue.task_done() > ``` > > It is
around the same area of code where I found the first issue that message A will
halt message B behavior. > I still question the design choice of letting the
exchange own the task rather than subscriber own the task > like the
"async.Queue()" does. > > Here is how the bug can be triggered. Let's say
producer a produce 3 message A1 A2 A3 at the same time > before the exchange
have any chance to run. The subscriber of this message on the channel should get
A1 > A2 A3, in that order. > 

Hey Chris -

I'm not seeing the ordering issue that you are calling out. Messages are
published through a FIFO queue, so the A1 A2 A3 ordering that you describe would
be preserved. Each task here is fanning out one message at a time in FIFO order
to all subscribers, so while subscribers B1, B2, or B3 may get notified of the
new `Message` in an undefined order through the `asyncio.gather()` call in
`_deliver_message_to_subscribers`, A1, A2, and A3 will be dispatched in the
order that they were enqueued.

But aside from the delivery semantics, I would strongly urge you not to bind
your designs tightly to Message ordering like this. In the OLAS and ServoX cases
before us, we can reasonably discuss and manage strictly ordered message
delivery. But as soon as we introduce multiprocessing to scale the Python
process horizontally or start to leverage networking and cluster nodes it
becomes impossible to define an absolute "correct" delivery order. Welcome to
the world of distributed systems. To build robust, future proof designs our
options are 1) Stop caring about time. Treat messages as standalone values where
delivery/processing order does not matter, 2) Use timestamps and hashing tricks
to synthesize an absolute order (e.g., utilize a monotonic `sequence` number,
detect when you have a gap in the sequence, and buffer/wait until the missing
data arrives), or 3) Utilize data structures that are resilient to out-of-order
delivery and deterministically converge to a final solution regardless of the
sequencing of the particular operations -- Operational Transforms are the most
commonly known flavor (this is what underlies Google Docs) but smart money is on
CRDTs (Conflict Free Replicated Data Type, see
https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type).

CRDTs have a
deterministic state given any subset of state transformations applied in any
order. Think about how a `Set` diverges from a `List` -- once you give up index
based operations (such as get item at index 3, insert item at index 1, or delete
3 items starting at index 2), you are left with a set of operations that are
safe to use concurrently without locks, semaphores, or mutexes. With a set, you can
add items, remove items, and check for the inclusion of items deterministically
regardless of the order in which the operations are applied. Once you let go of order
and indexing, the data becomes eventually consistent and distributed actors can mutate
it concurrently. It's so elegant and beautiful.

I have lived in all three of these worlds and I can tell you that while
everybody reaches for #2, the only truly viable paths are 1 and 3. Trying to
maintain a facade of reliable, strictly ordered eventing in a distributed system
is an exercise in madness.

I am not sure exactly what data you are moving around in OLAS land, but if you
send messages that contain complete object state (as opposed to
patches/mutations) and include a version number, TimeUUID, or timestamp (though
timestamps are also whackier than you would think because you need very closely
synchronized clocks and minimal drift to rely on it for distributed operations),
then you can compare the version/TimeUUID/timestamp to the last Message that you
processed and drop stale ones on the floor.

If you really need patches/mutable state, just bite the bullet and get wise on
CRDTs rather than trying to hack your way around reality.

> Here is the bug, because you create a new task for EACH message. There is
actually no guarantee the > subscriber's callback will execute the in the order
of A1 A2 A3. Because all 3 task are available. It is up to > the async scheduler
to schedule the 3 task at its choice. If A1 gets block, it is possible A3 get
execute and > complete before A2.

See above, you don't actually want this as much as you think you do.

> > As soon as you schedule 3 different task, you lose the ordering of the task.
That is why for each subscriber, > a separate async.Queue() is needed. Each
subscriber will have its own task, not each message. Since each > subscriber
need to have its own task, you might just let subscriber have that task.

Given a Subscriber consuming Messages from an arbitrary number of Publishers,
there is no such thing as a "correct" order. Don't produce or consume messages
that require strict ordering to succeed. You cannot actually guarantee ordering
or "once and only one" message delivery, so just don't even bother trying. It is
a design constraint.

> > BTW, this kind of bug is very subtle and very hard to debug if it actually
triggered. Most of the > time the system runs fine, only when it is buys and
every once a while it happen, and hard to reproduce.

It's not very subtle if you just make it impossible. Don't rely on absolute
ordering.

> > I still think expose the separate "async.Queue" to the subscriber is much
simpler and less boiler plate > code to expose a call back handler.

I think we have very different definitions of what constitutes boilerplate code.
I consider boilerplate code to be that which is required to satisfy ceremonial
but non-functional requirements -- its the stuff that is repeated across every
implementation without any differentiation. I think your are bristling against
the implementation code itself, not stuff that you are copying/pasting into the
application target?

I do not want to expose the asyncio.Queue object because it is an implementation
detail. We may change direction at any time. If you think about it from the
perspective of the Publisber and Subscriber roles, the Publisher wants to
broadcast information for all the world to see. The Subscriber wants to consume
a narrow set of Messages. Nobody care about the Queues.

> > Make the code simple and easy to understand is a good thing, it helps your
quest. > I don't see the conflict on that.

I assert that the code *is simple to understand*. You are conflating
understanding the down and dirty specifics of the implementation with the
abstractions and interface that is being enabled. Gavin, Lorenzo, and
Chris/Blake six months from now will not care nor be particularly interested in
how the `servo.pubsub` module is achieving its goals provided that it works and
behaves as described.

I stake the claim that the API usage examples explored in this gist:
https://gist.github.com/blakewatters/b7da891139210d3e8e959497d68e5b89 are very
simple to understand and crystal clear to the reader (provided that you speak
modern async Python, I suppose). The concepts that you want to think about and
program against are `Message`, `Channel`, `Publisher`, and `Subscriber` objects.
The fact that there are a couple of `asyncio.Queue` objects rattling around in
there is immaterial to anybody else but you and I because we are staring at
the implementation on this PR. If you can publish Messages and know that other
Connectors will receive them and cleanly subscribe to the Messages that you are
interested in and have confidence that you will receive them, you don't have to
care about exactly how it happens (setting aside egregious performance/security
violations).

Long response here, but this stuff is important and essential to understanding my
designs so I have tried to thoroughly unpack it.

Lastly, regarding the `I still question the design choice of letting the
exchange own the task rather than subscriber own the task` comment: I think you may
have learned why I went this direction the hard way with your invisible background
crashing bugs. Reality is that I know way more about asyncio and failure modes than
99% of the developers who will be working with pub/sub on a daily basis and the more
rope that you provide, the more likely it becomes that someone is going to hang
themself. Aside from the design/conceptual assertions made above (which I do believe
to be the stronger voice in this choir), by managing these tasks on behalf of the
developer I can take care of memory management and exception handling that most
developers will either ignore or fuck up trying to handle. The 103 (as of this writing)
tests in `tests/pubsub_test.py` let us all be lazy and professional at the same time.
I have bolted down handlers for every irritating quick, edge case, and failure mode that
I have encountered so that I can just never worry about them again. The lower level that
you go, the more of this stuff that you are taking direct responsibility for getting right.

PR #131, opened 2021-01-13
Feature/argo rollouts rebase SOL-42

linkous8:
Implementation of support for Argo Rollouts within existing Kubernetes connector logic

PR opened to trigger integration testing

### DIFF
--- a/tests/connectors/kubernetes_test.py
+++ b/tests/connectors/kubernetes_test.py
@@ -1110,6 +1110,7 @@ class TestKubernetesConnectorIntegration:
 ##
 # Rejection Tests using modified deployment
 @pytest.mark.integration
+@pytest.mark.clusterrolebinding('cluster-admin')
 @pytest.mark.usefixtures("kubernetes_asyncio_config")
 @pytest.mark.applymanifests("../manifests", files=["fiber-http-unready-cmd.yaml"])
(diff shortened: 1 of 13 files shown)


tests/connectors/kubernetes_test.py
```diff
@@ -1130,3 +1134,140 @@ async def test_adjust_never_ready(self, config, kube: kubetest.client.TestClient
         with pytest.raises(AdjustmentRejectedError):
             description = await connector.adjust([adjustment])
             debug(description)
+
+##
+# Tests against an ArgoCD rollout
+# @pytest.mark.clusterrolebinding('cluster-admin') # May be needed
```
blakewatters, line 1140: You are probably getting away without having this enabled because it's a cluster-wide resource and probably hanging out from another test -- this is a weakness of the current setup where we are reliant on a single reused cluster that exposes us to confirmation bias.

Still marinating on when and how to do it but maybe we switch to dynamically allocated clusters so there's better isolation.

There's also a nascent concept of `system` tests where I want to take a small number of end to end tests (e.g., do your full rollout adjust flow) but parametrize it across specific environments like Argo versions, GKE/EKS/AKS, different Kubernetes versions, etc. so we get a real compatibility matrix for the big-ticket functionality

linkous8:

PR #138, opened 2021-01-22
Feature/k8s resource limit min

linkous8:
Support for a new  `limit_min` configuration property of the CPU and Memory models. Only supported with `ResourceRequirements.compute` (equivalent of legacy `both` option). 

When set, the corresponding resource limit will only be set to the adjust value if said value is greater than the configured `limit_min`. Otherwise, limit will be set to the configured value of `limit_min`.

Note accommodating this change opened a bit of a can of worms as the adjust values were not being cast into comparable types prior to being fed into the containers. I'm hopeful any side effects are minor and can be quickly fixed but I am open to alternative implementations

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -2126,7 +2126,7 @@ class KubernetesAdjustment(servo.Adjustment):
         else:
             raise NotImplementedError(
                     f"missing error normalization logic for setting '{setting}'"
-                ) 
+                )
 
         for value in values:
             ret_values.append(parse_val(value))
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -2371,7 +2371,8 @@ class DeploymentOptimization(BaseOptimization):
         to enable aggregation of related adjustments and asynchronous application.
         """
         setting_name, value = adjustment.normalize_adjustment()
-        self.logger.info(f"adjusting {setting_name} to {value.human_readable()}")
+        log_val = value.human_readable() if isinstance(value, (Millicore, ShortByteSize)) else value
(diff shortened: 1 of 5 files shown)


servo/connectors/kubernetes.py
```diff
@@ -2022,8 +2022,20 @@ class CPU(servo.CPU):
     max: Millicore
     step: Millicore
     value: Optional[Millicore]
+    limit_min: Optional[Millicore] = pydantic.Field(
```
blakewatters, line 2025: I don't really love this name and this concept feels pretty shaky to me

blakewatters:
This overlaps with my flexible resources branch. Will look to lift and shift what we can during merge

blakewatters:
I'm having trouble getting down with this one conceptually.

It feels like we are attempting to apply a client-side fix to a server-side issue. Refusing to apply a prescribed change and coercing it into a different value strikes me as wrong and a bad idea.

servo/connectors/kubernetes.py
```diff
@@ -2053,6 +2065,11 @@ def validate(cls, v: pydantic.StrIntFloat) -> "ShortByteSize":
             v = v * GiB
         return super().validate(v)
 
+    def human_readable(self) -> str:
+        sup = super().human_readable()
+        if sup[-1] == 'B' and sup[-2].isalpha():
+            sup = sup[0:-1]
+        return sup
```
blakewatters, line 2072: Merging this change separate from the rest until we talk it out

linkous8:
Should be addressed by flexible resources changes. Failing that, @blakewatters has indicated it is more appropriate to apply this logic server side.

PR #142, opened 2021-02-01
fix: allow uppercase in app names

lswank:
@blakewatters Just give this a visual sanity check before sending it on. (Would have automagically merged, but you're my emotional safety net for the first few of these.)


blakewatters:
Apologies, there was actually more to be done here than just the casing.

Looking at the existing names that I could see, there were hyphens, underscores, and apparently we allow single character names. So I came over the top with a different fix.

PR #144, opened 2021-02-03
Lorenzo/sol 190 check that current pod request is within

blakewatters:


servo/connectors/opsani_dev.py
```diff
@@ -214,6 +214,40 @@ async def check_kubernetes_service_type(self) -> None:
                 f"expected service type of ClusterIP or LoadBalancer but found {service.spec.type}"
             )
 
+    @servo.checks.require("Target container resource requests are within limits")
```
blakewatters, line 217: Jump this up above the service checks to group container concerns together

servo/connectors/opsani_dev.py
```diff
@@ -214,6 +214,40 @@ async def check_kubernetes_service_type(self) -> None:
...
+    async def check_target_container_resources_within_limits(self) -> None:
+        # Load the Deployment
+        deployment = await servo.connectors.kubernetes.Deployment.read(
+            self.config.deployment,
+            self.config.namespace
+        )
+        # assert deployment, f"failed to read deployment '{self.config.deployment}' in namespace '{self.config.namespace}'"
```
blakewatters, line 224: Enable or drop this, not sure if its supposed to be here or gone

servo/connectors/opsani_dev.py
```diff
@@ -214,6 +214,40 @@ async def check_kubernetes_service_type(self) -> None:
...
+
+        # Find the target Container
+        target_container = None
+        for container in deployment.containers:
```
blakewatters, line 228: There's a `deployment.find_container()` method available to cut this down.

But for some functional one-liner foo to do the same you can do `target = next(filter(lambda c: c.name == self.config.container, deployment.containers), None)`

`next` returns the next value in an iterator, filter returns a new iterator after applying a predicate, and `next` accepts a default argument returned upon no values available.

servo/connectors/opsani_dev.py
```diff
@@ -214,6 +214,40 @@ async def check_kubernetes_service_type(self) -> None:
...
+            if container.name == self.config.container:
+                target_container = container
+                break
+
+        assert target_container, f"failed to find container '{self.config.container}' when verifying resource limits"
+
+        # Get resource requirements from container
+        container_cpu_request = servo.connectors.kubernetes.Millicore.parse(target_container.resources.requests["cpu"])
+        container_memory_request = servo.connectors.kubernetes.ShortByteSize.validate(target_container.resources.requests["memory"])
```
blakewatters, line 237: You can turn these into `@property` methods on the `Container` class so that the code reads `target_container.cpu` and `target_container.memory`

servo/connectors/opsani_dev.py
```diff
@@ -214,6 +214,40 @@ async def check_kubernetes_service_type(self) -> None:
...
+        config_cpu_min = self.config.cpu.min
+        config_cpu_max = self.config.cpu.max
+        config_memory_min = self.config.memory.min
+        config_memory_max = self.config.memory.min
+        
+        # Check values against config.
+        assert config_cpu_min <= container_cpu_request, f"target container requests '{container_cpu_request}' cpu but config specifies a minimum of '{config_cpu_min}'"
+        assert container_cpu_request <= config_cpu_max, f"target container requests '{container_cpu_request}' cpu but config specifies a maximum of '{config_cpu_max}'"
+        assert config_memory_min <= container_memory_request, f"target container requests '{container_memory_request}' memory but config specifies a minimum of '{config_memory_min}'"
+        assert container_memory_request <= config_memory_max, f"target container requests '{container_memory_request}' memory but config specifies a maximum of '{config_memory_max}'"
```
blakewatters, line 249: Extra credit here would be to switch this into a multi-check.

The value is that when they are run all checks that are not `required` are executed:

In a single check case like this, if you screwed up the CPU and the memory settings you would get one error (CPU), go fix it, then get the other error (memory), and go fix it.

With a multi-check expansion, they will both get run and you will see a progression -- this is why all the checks are so short and atomic.

tests/connectors/opsani_dev_test.py
```diff
@@ -109,6 +109,15 @@ async def test_resource_exists(
...
             assert result.success
 
+        async def test_target_container_resources_within_limits(
+            self, kube, checks: servo.connectors.opsani_dev.OpsaniDevChecks, config: servo.connectors.opsani_dev.OpsaniDevConfiguration
+        ) -> None:
+            config.cpu.min = "100m"
+            config.memory.min = "64MiB"
+            # TODO: Should explicity specify the max values as well as the target test values.
+            result = await checks.run_one(id=f"check_target_container_resources_within_limits")
+            assert result.success
```
blakewatters, line 119: Test failure case as well

servo/connectors/opsani_dev.py
lswank: Struggling with this. The concept is easy, but I am getting stuck in figure out the data types. This may be more of an IDE / visibility issue for me. May need help with the tooling.

linkous8: @lswank Here's the python [client model for ResourceRequirements](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1ResourceRequirements.md). As you can see, requests/limits are both dict[str, str] so I'm thinking your return type should be a string

Also @blakewatters shouldn't we also add property decorators `target_container.cpu_limit` and `target_container.memory_limit` for the sake of completeness?

PR #148, opened 2021-02-09
Commandline adjust invocations are not parsed/marshalled into usable units

linkous8:


servo/connectors/kubernetes.py
```diff
@@ -2185,17 +2195,32 @@ class ShortByteSize(pydantic.ByteSize):
...
+            v = float(v)
+
         if isinstance(v, str):
             try:
                 return super().validate(v)
             except:
                 # Append the byte suffix and retry parsing
                 return super().validate(v + "b")
-        elif isinstance(v, float):
+        elif isinstance(v, float): # TODO: shouldn't this apply to int as well?
```
blakewatters, line 2207: Been awhile since I've been in here but if I recall correctly the thinking was that an integer would specify an exact number of bytes whereas a float would be interpreted as being in `GiB` units. So `1` would be 1 byte and `1.0` would be 1 GiB.

Arguably I'm getting too cute with some of this shit but I think we have to think about it from configuration side looking in.

servo/utilities/subprocess.py
```diff
@@ -266,6 +266,10 @@ async def stream_subprocess_shell(
             f"Subprocess finished with return code {result} in {duration} (`{cmd}`)"
         )
         return result
+    except asyncio.CancelledError:
+        # when used in asynccontextmanager, the wrapping task is cancelled after yield (aexit), cleanup the subprocess in that case
+        process.terminate()
+        raise
```
blakewatters, line 272: Good catch here but I went with a slightly different spin on the handling: 76e6d2d47d95e0a6bbabbae76e75b2a108f0fdc1

tests/conftest.py
```diff
@@ -636,7 +637,7 @@ def _identifier_for_target(target: ForwardingTarget) -> str:
             ports_to_urls = dict(map(lambda p: (p[1], f"http://localhost:{p[0]}"), ports))
             yield ports_to_urls
     finally:
-        task.cancel()
+        task.cancel() if task else ...
```
blakewatters, line 640: Looks like I already fixed this one in a different way

blakewatters:
This one is gonna collide with the new resource branch also but looks like there may still be some good stuff to fold in post merge

linkous8:

PR #150, opened 2021-02-09
Lorenzo/sol 20 canary creation needs to report progress

lswank:
Draft pull request because I wanted to ask some questions.

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -37,6 +37,7 @@ import backoff
 import kubernetes_asyncio
 import kubernetes_asyncio.client
 import kubernetes_asyncio.client.exceptions
+import kubernetes_asyncio.watch
 import pydantic
 
 import servo
@@ -879,6 +880,10 @@ class Pod(KubernetesModel):
         self.logger.trace(f"pod: {self.obj}")
 
         async with self.preferred_client() as api_client:
+            async with kubernetes_asyncio.watch.Watch().stream(api_client.list_pod_for_all_namespaces) as stream:
+                async for event in stream:
+                    evt, obj = event['type'], event['object']
+                    self.logger.info(f'"{evt}" pod "{obj.metadata.name}" in NS "{obj.metadata,namespace}"')
             self.obj = await api_client.create_namespaced_pod(
                 namespace=namespace,
                 body=self.obj,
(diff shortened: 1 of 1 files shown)


servo/connectors/kubernetes.py
```diff
@@ -875,10 +876,18 @@ async def create(self, namespace: str = None) -> None:
         if namespace is None:
             namespace = self.namespace
 
-        self.logger.info(f'creating pod "{self.name}" in namespace "{self.namespace}"')
+        self.logger.info(f'Attempting to create a pod named {self.name} in the namespace {self.namespace}.')
         self.logger.trace(f"pod: {self.obj}")
 
         async with self.preferred_client() as api_client:
+            async with kubernetes_asyncio.watch.Watch().stream(api_client.list_pod_for_all_namespaces) as stream:
```
lswank, line 883: Since this is my first attempt at using the Kubernetes API, I've taken what I am sure is a very naieve approach to grabbing pod data. I'm sure this is subscribing us to all sorts of data that we don't want or need. If this is fine for the time being, I can turn this from a draft request into an actual PR.

servo/connectors/kubernetes.py
```diff
@@ -875,10 +876,18 @@ async def create(self, namespace: str = None) -> None:
...
+                async for event in stream:
+                    event_type, event_object = event['type'], event['object']
+                    # TODO: Really should find a way to select in the API call so that we aren't registering for all events for all time.
+                    # This will be frowned upon in permissive environments where our API calls will return many more pods.
+                    if event_object.metadata.namespace == self.namespace:
+                        self.logger.info(f'Kubernetes reports that we successfully {event_type.lower()} a pod named {event_object.metadata.name} in the {event_object.metadata.namespace} namespace.')
```
lswank, line 889: For pretty formatting purposes I've gone ahead and done some string manipulation, but every part of me is saying, "You can't be guaranteed that Kubernetes API call results are strings. What are you even thinking? That's angrily coercing data types in a way that isn't even close to explicit." And yet another part of me goes, "I bet that's just the Python way of doing things."

Thoughts?

linkous8: Watch event's `type` field is a [str in the backing model](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1WatchEvent.md), so it should be guaranteed to be a string by the protocol buffers

PR #151, opened 2021-02-10
Fred/sol 46 kubernetes bad request errors are not

linkous8:


blakewatters:
Decorator implementation looks a little funky. Couldn't we move this off class into a standalone function?

linkous8:
Thanks for the feedback, I've factored it into a new implementation off the latest main since merge generated a ton of conflicts and `git rebase` is poor form on open PRs.

While running the tests, I noticed something I hadn't caught before in that the reference object for the baseline deployment is being updated during canary adjustment due to a shallow copy leaving nested references in tact. This fix probably has some overlap with your flexible resources branch as I noticed you did some significant restructuring of the CanaryOptimization class in it. However, my fix is very small and should be a good interim fix until that branch is landed

See the new PR here: https://github.com/opsani/servox/pull/197

PR #154, opened 2021-02-12
Fix overriding of subconfig timeout SOL-220

### LINKED ISSUE #153: Kubernetes top level config timeout can override subconfig timeouts
Apologies as I introduced this one myself in implementing the rejection logic. I made the mistake of [setting and using this timeout property](https://github.com/opsani/servox/blob/464a2e5672921ff2a70dda5f5b951785ebdb99a5/servo/connectors/kubernetes.py#L2260) instead of referencing [this one](https://github.com/opsani/servox/blob/464a2e5672921ff2a70dda5f5b951785ebdb99a5/servo/connectors/kubernetes.py#L3197) which is cascaded down to the deployment config level.

This could result in undesirable behavior in cases where we do not expect the timeout to be cascaded from the top level of the config. See incoming PR for fix

linkous8:
Resolves #153 

* Reference configs instead of self.timeout
* Don't pass parent config timeout to child config obj create

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -2465,7 +2465,7 @@ class DeploymentOptimization(BaseOptimization):
         self.logger.info(f"adjustment failed: rolling back deployment... ({error})")
         await asyncio.wait_for(
             asyncio.gather(self.deployment.rollback()),
-            timeout=self.timeout.total_seconds(),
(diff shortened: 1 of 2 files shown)


linkous8:
SOL-220

blakewatters:
I have superceded this with changes on the flexible resources branch that is about to land.

PR #155, opened 2021-02-12
Add basic check for node allocatable resources

linkous8:
* Check will fail if any schedulable nodes have < 512Mib and 500mCPU
* Add check failure test coverage


blakewatters:
What's the genesis of this one? Looks pretty close to merge ready but this one wasn't on my radar (or slipped my mind)

linkous8:
Per the associated issue linked by the linear-app bot in the comment above, you said you'd run into problems caused by funky nodes with bad performance.

PRed functionality is a naive attempt at scanning for such instances but LMK if I can do anything to dial it in further

blakewatters:
I'm worried about merging this in the context of a larger cluster.

I think to do it right we'll have to consider shit like node selectors, affinity, and anti-affinity which can get pretty squirrely.

servo/connectors/kubernetes.py
```diff
@@ -3324,6 +3324,8 @@ async def load_kubeconfig(self) -> None:
 DeploymentOptimization.update_forward_refs()
 CanaryOptimization.update_forward_refs()
 
+NODE_MIN_CPU = 500
+NODE_MIN_MEM = 512 * MiB
```
blakewatters, line 3328: If we want to land this we should prolly push this into the configuration class so that we have an escape hatch

linkous8:
> I'm worried about merging this in the context of a larger cluster.
> 
> I think to do it right we'll have to consider shit like node selectors, affinity, and anti-affinity which can get pretty squirrely.

Thats a good point. I mostly opened this PR as a starting point to get discussion going and don't see much value add from landing it in the current state. Closing for now, will update the Linear task discussion accordingly

PR #156, opened 2021-02-16
Cleanup Kubernetes wait log output

linkous8:


blakewatters:
I think we want this at info level

PR #157, opened 2021-02-19
Add hints for step aligned min/max, allow step_alignment to work with offset min

linkous8:
- change order of parsing step and max so max can validate against step
-  add validation error if step is greater than max
- update step alignment validation to account for offset min
- add hints for closes min max values if not step aligned


linkous8:
Resolved by https://github.com/opsani/servox/commit/c96b1ba1a8613d74ad99bae80a6a8298038d54ba#

PR #159, opened 2021-02-23
Implement User Agent string to provide environment info and reduce data blindness

linkous8:
- Remove hard coded user_agent from api
- Add dynamic user agent handling to optimizer class
- Populate ua platform of servo and namespace if available
- connector.py add missing import servo.configuration
- set api_client_options to use optimizer user agent
- populate/remove k8s user agent info on attach/detach


servo/connectors/kubernetes.py
```diff
@@ -3564,6 +3564,23 @@ async def attach(self, servo_: servo.Servo) -> None:
...
+            version = "UnknownVersion"
+            comment_dict = { "namespace": self.config.namespace }
+            async with kubernetes_asyncio.client.api_client.ApiClient() as api:
+                v1 =kubernetes_asyncio.client.VersionApi(api)
+                version_obj = await v1.get_code()
+                version = f"{version_obj.major}.{version_obj.minor}"
+                comment_dict["platform"] = version_obj.platform
+
+            comment_str = "; ".join(list(map(lambda tup: f"{tup[0]} {tup[1]}", comment_dict.items())))
+            self.optimizer.add_user_agent(self.name, f"{version} ({comment_str})")
```
blakewatters, line 3578: Maybe we use the shared config cascading stuff to expose `environment` or `telemetry` so that all connectors can do `self.telemetry['kubernetes.version'] = ...`

We'll also want to get some data from other connectors like the Prometheus like the Prometheus version we are talkiing to.

blakewatters (changes requested):
Sorry for how long this has been hanging in the air. I think if we can just improve the factoring design a bit we can land it and then start progressively introducing more and more telemetry data.

This is a good foundation but we're not quite there on the design yet.

linkous8:
Closing in favor of refactor

PR #180, opened 2021-03-23
Incorporate settlement handling into EventProgress

linkous8:
- scale settlement progress to second half of range to avoid reset to 0
- defer to timeout progress if further along to keep finalization logic
- update kubernetes settlement handling use EventProgress


linkous8:
PR has been superseded by numerous settlement fixes/changes some of which are still in progress. Closing in favor of a refactor once settlement logic is in a more stable place

PR #189, opened 2021-04-07
Fred/sol 174 detect configmap changes and restart

linkous8:
Establish watch on configuration servo.yaml and shutdown Servo when updated


servo/runner.py
```diff
@@ -389,7 +412,7 @@ def _display_banner(self) -> None:
                      'MAGENTA': colorama.Fore.MAGENTA, 'CYAN': colorama.Fore.CYAN,
                      'RAINBOW': colorama.Fore.MAGENTA}
 
-        terminal_size = os.get_terminal_size()
+        terminal_size = shutil.get_terminal_size()
```
blakewatters, line 415: Merged this. Wasn't aware of it. Nice catch.

servo/runner.py
```diff
@@ -490,12 +513,15 @@ async def _shutdown(self, loop, signal=None):
...
             self.logger.critical(f"Failed assembly shutdown with error: {error}")
+            self.logger.opt(exception=error).debug("Failed assembly shutdown with error")
 
         await asyncio.gather(self.progress_handler.shutdown(), return_exceptions=True)
         self.logger.remove(self.progress_handler_id)
 
         # Cancel any outstanding tasks -- under a clean, graceful shutdown this list will be empty
         # The shutdown of the assembly and the servo should clean up its tasks
+        # TODO: ServoRunner.shutdown() does not clean up tasks, only sets its main_loop LCV to false
+        #   but the loops, in most cases, won't complete another iteration before the logic below shuts it down forcefully
```
blakewatters, line 524: Not following fully -- what's LCV?

tests/assembly_test.py
```diff
@@ -0,0 +1,94 @@
...
+        config_file=servo_yaml, optimizer=optimizer
+    ))
+    return assembly_
+
+@pytest.fixture
+def assembly_runner(assembly: servo.Assembly) -> servo.runner.AssemblyRunner:
+    """Return an unstarted assembly runner."""
+    return servo.runner.AssemblyRunner(assembly)
+
+@pytest.mark.timeout(10)
```
blakewatters, line 67: I jettisoned `pytest-timeout` because it has some super fun non-deterministic conflicts with `pytest-xdist` and I value the fast test suite more highly

servo/runner.py
```diff
@@ -368,6 +370,27 @@ def handle_progress_exception(error: Exception) -> None:
...
+                    change = next(filter(lambda c: c[1] == str(self.assembly.config_file), changes), None)
+                    if not change:
+                        continue
+                else:
+                    change = changes.pop()
+
+                self.logger.critical(f"Config file change detected ({str(change[0])}), shutting down active Servo(s) for config reload")
+                self.logger.trace(f"Config file watch change: {change}")
+
+                loop.create_task(self._shutdown(loop))
```
blakewatters, line 389: I think we can do better here without too much effort.

We should be able to take the existing config and diff it against the incoming config and selectively restart the connectors that have had their configuration changed. We have a known mapping of configuration keys to in-memory objects.

I think we can even add/remove connectors and it shouldn't be that hard because we really only need to distill changes down to the top-level key affected.

I am very interested in this because I think where we want to go is clustering servos to provide HA, seamless upgrades, and keeping our resource utilization constrained at scale... I had 5000 servos running in one assembly back in October. Imagine if we just split that up across 5 instances with a gossip protocol paired with a Kubernetes mutating webhook admission controller to optimize everything as it is deployed

servo/runner.py
linkous8: Loop Control Variable

linkous8:
Per Blake's review, a different approach is needed to updates detected in the configuration. Closing this in favor of a new PR

PR #195, opened 2021-04-11
Added check for container minimum request with no

blakewatters:
headroom.


blakewatters:
This needs to be updated to use the new properties that respect the `get` config

PR #197, opened 2021-04-12
Fix handle_error destroy logic fails on non-existent resources (Fixes SOL-46)

linkous8:
Rebase stale changes and factor in PR feedback from https://github.com/opsani/servox/pull/151
- 404 delete implicit decorator changed to explicit stand alone function
- Fixed baseline unintenionally updated due to shallow copy
- skip wait_until_deleted if tuning pod does not exist
- raise handled tuning errors so they are logged in the backend
- added test to invoke destroy error handling on non-existent tuning pod


blakewatters:
I think we can factor this up onto the base class. There's some similar stuff I am thinking about around how to handle 409 conflicts.

Merging this down as is since you have it nicely pinned down with a test but I think we're gonna want to make a big pass through to reconcile these cases and tighten it all up.

PR #200, opened 2021-04-13
Fix raise_for_status bugs and CanaryOptimization.destroy not clearing tuning pod (Fixes SOL-308, SOL-342)

linkous8:
- fix usage of async done callback _raise_for_task which will never run
- add settlement failed test for reproducing bugs above against main
- refactor rejection test for flexibility and reduce redundant manifests
- fix CanaryOptimization.destroy not clearing tuning pod


servo/connectors/kubernetes.py
```diff
@@ -3009,25 +3010,30 @@ async def apply(self, adjustments: List[servo.Adjustment]) -> None:
...
-        for future in asyncio.as_completed(tasks, timeout=self.config.timeout):
+        for future in asyncio.as_completed(tasks, timeout=self.config.timeout.total_seconds()):
             try:
                 await future
             except Exception as error:
                 servo.logger.exception(f"Optimization failed with error: {error}")
 
+        # TODO: first handler to raise will likely interrupt other tasks.
+        #   Gather with return_exceptions=True and aggregate resulting exceptions before raising
+        await asyncio.gather(*handle_error_tasks)
```
blakewatters, line 3035: Why do we want to gather this instead of just letting it run?

servo/connectors/kubernetes.py
```diff
@@ -3521,7 +3527,7 @@ async def readiness_monitor() -> None:
                 while not progress.finished:
                     if not await state.is_ready():
                         # Raise a specific exception if the optimization defines one
-                        state.raise_for_status()
+                        await state.raise_for_status()
```
blakewatters, line 3530: Went ahead and merged this change as a one-off

servo/connectors/kubernetes.py
linkous8: 1. We want to block flow of control so that rollbacks and teardowns are allowed to complete before reporting results to the backend
2. The handle_error task is responsible for communicating the error its handling to the backend. If we allow said tasks to run in the background, the connector would likely [fall through to a more generic error](https://github.com/opsani/servox/blob/faea00e0a6557cd8d7beb0852614d1616e44aef5/servo/connectors/kubernetes.py#L3597) which is undesirable
3. I'm pretty sure exceptions raised by the [handle_error tasks](https://github.com/opsani/servox/blob/faea00e0a6557cd8d7beb0852614d1616e44aef5/servo/connectors/kubernetes.py#L3081) would not be retrieved which would make for some messy logs as well

rstarmer (approved):
I've reviewed the changes, and this code looks good to me.

servo/connectors/kubernetes.py
```diff
@@ -3103,6 +3098,10 @@ async def _raise_for_task(optimization: BaseOptimization, task: asyncio.Task) ->
             except Exception as error:
                 servo.logger.exception(f"Optimization failed with error: {error}")
 
+        # TODO: first handler to raise will likely interrupt other tasks.
```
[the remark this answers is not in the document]
linkous8: Sorry I missed this, its being tracked under SOL-311

PR #201, opened 2021-04-13
Fix target_deployment updated during canary apply

linkous8:
- Add test to demonstrate bug against main and guard regression


blakewatters:
Getting fancy here but I'll allow it. I got burned on a similar `copy` vs `deepcopy` issue end of last week

This is the reason why I've heavily favored more transient/stateless implementations. But these Kubernetes resources are so deeply nested that it's impractical.

Maybe the lesson is to default to using `deepcopy` until it becomes a performance issue?

blakewatters:
@linkous8 Throw eyes on this. I think we've dealt with the same issue in different ways but if I am missing some nuance from your changes I want to make sure that we capture them

PR #202, opened 2021-04-14
Fix Kubernetes deployment model using spec.selector.match_labels to locate itself

linkous8:


blakewatters:
Another great catch here.

Definitely just got lucky off of naming conventions but this was just totally wrong. Maybe we need some fuzzing for some of this shit?

PR #204, opened 2021-04-14
Fix kubernetes deployment readiness check returning false positives

linkous8:


blakewatters:
Another great catch. Merging

PR #208, opened 2021-04-16
Update optimization classes and handle_error definitions to use nested configuration

linkous8:


linkous8:
Opening fresh PR to see if that will resolve CI tests

PR #213, opened 2021-04-19
Update optimization classes and handle_error definitions to use nested on_failure configuration

linkous8:


blakewatters:
This one I am not fully understanding.

Where is the cascade coming from? It looks like the properties are hard-wired to the deployment config. Maybe this is just a commenting issue or I am getting tired but I can't follow it all the way through

servo/connectors/kubernetes.py
```diff
@@ -2813,7 +2839,10 @@ async def destroy(self, error: Optional[Exception] = None) -> None:
 
         self.logger.info(f'destroyed tuning Pod "{self.name}"')
 
-    async def handle_error(self, error: Exception, mode: "FailureMode") -> bool:
+    async def handle_error(self, error: Exception, mode: "FailureMode" = None) -> bool:
+        if mode is None:
+            mode = self.on_failure
+
```
blakewatters, line 2845: Why didn't you just yank the `mode` param entirely?

linkous8:
> This one I am not fully understanding.
> 
> Where is the cascade coming from? It looks like the properties are hard-wired to the deployment config. Maybe this is just a commenting issue or I am getting tired but I can't follow it all the way through

`on_failure` [is defined on the BaseKubernetesConfiguration class](https://github.com/opsani/servox/blob/d5687d05de05d67a2753f7cfa483608914e371f9/servo/connectors/kubernetes.py#L3312) which both KubernetesConfiguration and DeploymentConfiguration inherit from. The `cascade_common_settings` [method of KubernetesConfiguration](https://github.com/opsani/servox/blob/d5687d05de05d67a2753f7cfa483608914e371f9/servo/connectors/kubernetes.py#L3379) called during config init cascades the value for this setting down to the deployment configuration as demonstrated by the following log from some quick testing I did to make sure it was working the way I thought it was:

```
2021-04-26 09:56:18.019 | TRACE    | servo.connectors.kubernetes:cascade_common_settings:3368 | servo - cascaded setting 'on_failure' from KubernetesConfiguration to child '[DeploymentConfiguration(description=None, kubeconfig=None, context=None, namespace='kubetest-test-adjust-handle-error-respects-nested-co-1619448977', settlement=None, on_failure=<FailureMode.exception: 'exception'>, timeout=None, name='fiber-http', containers=[ContainerConfiguration(description=None, name='fiber-http', alias=None, command=None, cpu=CPU(name='cpu', type='range', pinned=False, value=None, min=125, max=875, step=125, requirements=<ResourceRequirements.compute: 3>), memory=Memory(name='mem', type='range', pinned=False, value=None, min=134217728, max=805306368, step=33554432, requirements=<ResourceRequirements.compute: 3>), env=None)], strategy=<OptimizationStrategy.default: 'default'>, replicas=Replicas(name='replicas', type='range', pinned=False, value=None, min=1, max=4, step=1))]': value=exception
```

servo/connectors/kubernetes.py
linkous8: I thought it would allow greater flexibility, I'll push an update to remove it

rstarmer (approved):
Seems that this was a resolution of a previous review.

LGTM

rstarmer (approved):
This looks good to me.

PR #216, opened 2021-04-22
Fix Assembly.shutdown gathering booleans instead of shutdown async methods when any assembled servos are already shut down

linkous8:
Updates the Assembly.shutdown method to filter running Servos prior to mapping out their respective shutdown methods


blakewatters:
Great catch.

I just merged an update to `pytest-asyncio` that smells like it may cover what you called out in the comments but we have bigger fish to fry so I'm just merging as is but FYI

blakewatters:
Moved this into ready for review because it looked solid to me.

PR #219, opened 2021-04-26
Update failing test to account for changes in main

linkous8:
- mock patch disable memory assignment validation
- Update log message being verified
- Improve test documentation with comment


linkous8:
Moving to new branch with fix for other integration test failure `test_process`

PR #226, opened 2021-05-06
Fix optimizer token logged in plain text when level set to TRACE

linkous8:


linkous8:
Closing for rebase with cleaner test runs

PR #228, opened 2021-05-07
Ensure main_replicas max is static under HPA

linkous8:
- Add validation to indicate DeploymentOptimization incompatible with
 horizontal pod autoscaler
- Add config option to indicate deployment is controlled by an HPA
 - Report main_replicas max as maxint if mainline deployment has HPA


linkous8:
This change has been superseded by temporary patch (#228). Closing this as the long term solution will likely use a different approach

PR #229, opened 2021-05-07
Fred/sol 335 integration tests

linkous8:
Fix integration tests failing against main branch

### DIFF
--- a/tests/connectors/opsani_dev_test.py
+++ b/tests/connectors/opsani_dev_test.py
@@ -500,8 +500,12 @@ class TestServiceMultiport:
 
                 # Step 7
                 servo.logger.critical("Step 7 - Bring tuning Pod online")
-                async with change_to_resource(deployment):
-                    await deployment.create_or_recreate_tuning_pod()
+                # TODO: why is the tuning pod being created here when the check will recreate it anyway?
+                kubernetes_config = checks.config.generate_kubernetes_config()
+                canary_opt = await servo.connectors.kubernetes.CanaryOptimization.create(
+                    deployment_config=kubernetes_config.deployments[0], timeout=kubernetes_config.timeout
+                )
+                await canary_opt.create_tuning_pod()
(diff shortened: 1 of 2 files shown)


tests/connectors/opsani_dev_test.py
```diff
@@ -500,17 +500,27 @@ async def wait_for_targets_to_be_scraped() -> List[servo.connectors.prometheus.A
...
                 servo.logger.critical("Step 8 - Verify Service traffic makes it through Envoy and gets aggregated by Prometheus")
                 async with kube_port_forward(f"service/fiber-http", port) as service_url:
                     await load_generator(service_url).run_until(wait_for_targets_to_be_scraped())
 
+                # NOTE it can take more than 2 scrapes before the tuning pod shows up in the targets
+                scrapes_remaining = 3
                 targets = await wait_for_targets_to_be_scraped()
-                assert len(targets) == 2
+                while len(targets) != 3 and scrapes_remaining > 0:
+                    targets = await wait_for_targets_to_be_scraped()
```
lswank, line 520: Is there another Kubernetes API event handler we could be listening to so that we don't have to loop like this? (I don't know that there is, but if there is, it might be more graceful to create a wait_for_appearance_of_scraped_targets(target_count=3) or something. (Not a blocker. Just worth exploring.)

tests/connectors/opsani_dev_test.py
```diff
@@ -500,17 +500,27 @@ async def wait_for_targets_to_be_scraped() -> List[servo.connectors.prometheus.A
 
                 # Step 7
                 servo.logger.critical("Step 7 - Bring tuning Pod online")
-                async with change_to_resource(deployment):
-                    await deployment.create_or_recreate_tuning_pod()
+                # TODO: why is the tuning pod being created here when the check will recreate it anyway?
```
lswank, line 503: Isn't this the difference between create_tuning_pod() and create_or_recreate_tuning_pod()?

PR #239, opened 2021-05-12
Fix optimizer token logged in plain text when level set to TRACE

linkous8:


rstarmer (approved):
This looks like a staight forward way of redacting the token in log files.

LGTM

DanielHHowell (approved):
Looks solid

PR #240, opened 2021-05-19
Support setlement in opsani_dev (Fixes SOL-346, SOL-352)

linkous8:
- Fix control not included in adjust event dispatch
- Add opsani_dev config to allow override of timeout and settlement


rstarmer (approved):
This looks good to me and has added tests to verify behavior.

PR #247, opened 2021-05-26
Fix k8s connector's incorrect usage of resource_version in get latest pods uses

linkous8:


servo/connectors/kubernetes.py
```diff
@@ -2906,7 +2906,7 @@ async def handle_error(self, error: Exception) -> bool:
                         f"cannot rollback a tuning Pod: falling back to destroy: {error}"
                     )
 
-                await asyncio.wait_for(self.destroy(), timeout=self.timeout.total_seconds())
+                await asyncio.wait_for(self.destroy(), timeout=self.timeout.total_seconds() + 60)
```
rstarmer, line 2909: @linkous8 Is this to avoid 0 timeout?

rstarmer (approved):
I would like to clarify the one comment, but this looks good to me.

servo/connectors/kubernetes.py
```diff
@@ -1487,7 +1487,7 @@ async def get_latest_pods(self) -> List[Pod]:
                 for ownRef in rs.metadata.owner_references)]
         if not rs_list:
             raise servo.ConnectorError('Unable to locate replicaset(s) for deployment "{self.name}"')
-        latest_rs = sorted(rs_list, key= lambda rs: rs.metadata.resource_version, reverse=True)[0]
+        latest_rs = sorted(rs_list, key= lambda rs: int(rs.metadata.annotations['deployment.kubernetes.io/revision']), reverse=True)[0]
```
blakewatters, line 1490: Can we use `observed_generation` instead of going to the annotations like this? (Refs: https://stackoverflow.com/questions/47100389/what-is-the-difference-between-a-resourceversion-and-a-generation, https://thenewstack.io/kubernetes-deployments-work/)

They should be equal to the same value other than being in a string on the annotation vs. an `int` but it feels more idiomatic/maintainable to utilize an API model affordance versus mining shit out of the annotations

servo/connectors/kubernetes.py
blakewatters: Yeah what's up with the + 60 here?

tests/kubernetes_test.py
```diff
@@ -390,6 +390,29 @@ async def test_patch_service(self, kube: kubetest.client.TestClient) -> None:
...
+    _, old_rset = kube.get_replicasets().popitem()
+
+    # Generate a new replicaset
+    kube_dep = kube.get_deployments()["fiber-http"]
+    kube_dep.obj.spec.template.spec.containers[0].resources.requests["memory"] = "128Mi"
+    kube_dep.api_client.patch_namespaced_deployment(kube_dep.name, kube_dep.namespace, kube_dep.obj)
+
+    servo_dep = await servo.connectors.kubernetes.Deployment.read("fiber-http", kube.namespace)
+    # Timing is a bit tricky on this one
+    await asyncio.sleep(1.5)
```
blakewatters, line 406: Is there an event we can watch for or a value we can poll? Every time we blindly use sleep values like this the damned test winds up becoming brittle later on and I have to go back and hammer on it again

servo/connectors/kubernetes.py
linkous8: The canary pod destroy was timing out consistently even though the timeout for the test is 20 seconds. I didn't want to extend the config timeout as that will just make the test run time longer so I added a bit of extra buffer for it since its error handling logic.

Would you prefer the configured timeout be extended or should we perhaps remove the timeout from the handle_error ops?

servo/connectors/kubernetes.py
pnickolov: We should be able to handle the case when this annotation is missing - I don't think there is a guarantee it will be there. In any case, at least some protection should be avalable (even if it doesn't work, it can fail gracefully and indicate the error)

linkous8: Unfortunately that will not work in the context of replicasets. observedGeneration is equal to the deployment revision **on the Deployment object** but this does not hold true for the ReplicaSet object.

From https://thenewstack.io/kubernetes-deployments-work/

> As you can see, the ReplicaSet has indeed the label run=my-nginx. Not only that, the metadata contains again the deployment revision number. This is how a Deployment knows which ReplicaSet is the current one.

Further you can look at the yaml from this issue comment and see that the deployment.kubernetes.io/revision does not match the generation or observedGeneration of the replicasets

https://github.com/kubernetes/kubernetes/issues/41641#issuecomment-281064331

blakewatters: Ack. I'm not fully up to speed on `ReplicaSet` and `Deployment` nuance yet.

So the `ReplicaSetStatus` object doesn't align with the annotation?

tests/kubernetes_test.py
blakewatters: I'm looking bottom's up from the code at:

* https://github.com/tomplus/kubernetes_asyncio/blob/master/kubernetes_asyncio/client/models/v1_replica_set.py
* https://github.com/tomplus/kubernetes_asyncio/blob/master/kubernetes_asyncio/client/models/v1_replica_set_spec.py
* https://github.com/tomplus/kubernetes_asyncio/blob/master/kubernetes_asyncio/client/models/v1_replica_set_status.py

servo/connectors/kubernetes.py
linkous8: No it does not

blakewatters: Okay, I'm cool to merge then.

Are there cases where the annotation might not exist? Like should we defensively use `rs.metadata.annotations.get('deployment.kubernetes.io/revision', 0)`?

servo/connectors/kubernetes.py
blakewatters: I don't think that removing the timeout is safe -- I don't want us waiting forever. Unless maybe there's already a higher-order timeout in effect? I've tried to unwind nested overlapping timeouts but maybe this is another such case that I missed.

I really don't like the arbitrary slop -- it seems like something is off target

blakewatters: It's really taking more than 20s to destroy this Pod?

servo/connectors/kubernetes.py
linkous8: That annotation should be guaranteed to exist on any replicaset managed by a deployment. Returning 0 as a default could lead to some inconsistent and difficult to debug behavior so I am opting to fail gracefully per Peter's request:

```
        if missing_revision_rsets := filter(lambda rs: 'deployment.kubernetes.io/revision' not in rs.metadata.annotations, rs_list):
            raise servo.ConnectorError(
                f'Unable to determine latest replicaset for deployment "{self.name}" due to missing revision annotation in replicaset(s)'
                f' "{", ".join(list(map(lambda rs: rs.metadata.name, missing_revision_rsets)))}"'
            )
```

servo/connectors/kubernetes.py
linkous8: Apparently so. You can see it in this failed test run: https://github.com/opsani/servox/runs/2677588629

```
fut = <Task cancelled name='Task-464' coro=<CanaryOptimization.destroy() done, defined at /home/runner/work/servox/servox/servo/connectors/kubernetes.py:2893>>
timeout = 20.0

...

>               raise exceptions.TimeoutError()
E               asyncio.exceptions.TimeoutError
```

servo/connectors/kubernetes.py
blakewatters: Boom. I'm onboard with this approach

servo/connectors/kubernetes.py
blakewatters: What about just boosting the timeout on the config within your test body?

linkous8: Done. (May need to boost it further if this next test still fails)

tests/kubernetes_test.py
linkous8: Sure thing: https://github.com/opsani/servox/blob/fred/sol-319-k8s-connector-get_latest_pods-uses/tests/kubernetes_test.py#L406-L409

rstarmer (approved):
LGTM.

PR #258, opened 2021-06-04
Implement correct rejection reasons (Fixes SOL-344, SOL-349)

linkous8:
- Add tuning optimization check for restarted containers
- Update scheduling-failed reasons to unschedulable
- Don't treat start-failed as unschedulable
- update event_type == ERROR to use message instead of reason
- Update ReplicaFailure, Progressing.status == False to correct reason
- Move handle_error out of apply timeout error handling to prevent
 redundant invokation
- wait for gather awaitable cancellation on timeout to prevent
 exception never retrieved warning
- Set correct message, reason on final settlement check, add warning that
 handle_error was not run for it
- Refactor tests to provide context on the error if it fails assertions
- Add tests for OOMKilled on adjustment
- Add test for tuning optimization start-failed

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -910,11 +910,28 @@ class Pod(KubernetesModel):
         if not status.conditions:
             raise RuntimeError(f'Pod is not running: {self.name}')
 
+        restarted_container_statuses = list(filter(lambda cont_stat: cont_stat.restart_count > 0, (status.container_statuses or [])))
+        if restarted_container_statuses:
+            container_messages = list(map(lambda cont_stat: f"{cont_stat.name} x{cont_stat.restart_count}", restarted_container_statuses))
+            raise servo.AdjustmentRejectedError(
+                f"Tuning optimization {self.name} crash restart detected on container(s): {', '.join(container_messages)}",
+                reason="unstable"
+            )
+
         self.logger.trace(f"checking status conditions {status.conditions}")
         for cond in status.conditions:
-            if cond.reason in {"Unschedulable", "ContainersNotReady"}:
+            if cond.reason == "Unschedulable":
                 # FIXME: The servo rejected error should be raised further out. This should be a generic scheduling error
-                raise servo.AdjustmentRejectedError(message=cond.message, reason="scheduling-failed")
+                raise servo.AdjustmentRejectedError(
+                    message=cond.message,
+                    reason="unschedulable"
+                )
+
+            if cond.type == "Ready" and cond.status == "False":
+                raise servo.AdjustmentRejectedError(
+                    message=f"(reason {cond.reason}) {cond.message}",
+                    reason="start-failed"
+                )
 
             # we only care about the condition type 'ready'
             if cond.type.lower() != "ready":
@@ -1786,7 +1803,7 @@ class Deployment(KubernetesModel):
                     if event_type == "ERROR":
                         stream.stop()
                         # FIXME: Not sure what types we expect here
-                        raise servo.AdjustmentRejectedError(reason=str(deployment))
+                        raise servo.AdjustmentRejectedError(str(deployment))
 
                     # Check that the conditions aren't reporting a failure
                     if status.conditions:
@@ -1845,8 +1862,7 @@ class Deployment(KubernetesModel):
                 # TODO: Check what this error looks like
                 raise servo.AdjustmentRejectedError(
                     "ReplicaFailure: message='{condition.status.message}', reason='{condition.status.reason}'",
-                    condition.status.message,
-                    reason=condition.status.reason
+                    reason="start-failed"
                 )
 
             elif condition.type == "Progressing":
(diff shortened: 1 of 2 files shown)


servo/connectors/kubernetes.py
```diff
@@ -3125,29 +3140,34 @@ async def apply(self, adjustments: List[servo.Adjustment]) -> None:
...
+                    return_exceptions=True,
                 )
-
-                for result in results:
-                    if isinstance(result, Exception):
-                        for optimization in self.optimizations:
-                            if await optimization.handle_error(result):
-                                # Stop error propagation once it has been handled
-                                break
+                results = await asyncio.wait_for(gather_apply, timeout=timeout.total_seconds() + 60) # allow sub-optimization timeouts to expire first
```
rstarmer, line 3147: Curious about the need for a static value for additional time.  If this is a parameter, should we set it to a default and let the end user break things if they set it too short?

servo/connectors/kubernetes.py
```diff
@@ -1845,8 +1862,7 @@ def _check_conditions(self, conditions: List[kubernetes_asyncio.client.V1Deploym
                 # TODO: Check what this error looks like
                 raise servo.AdjustmentRejectedError(
                     "ReplicaFailure: message='{condition.status.message}', reason='{condition.status.reason}'",
```
pnickolov, line 1864: needs to be an f-string

servo/connectors/kubernetes.py
```diff
@@ -1786,7 +1803,7 @@ async def rollout(self, *, timeout: Optional[servo.DurationDescriptor] = None) -
                     if event_type == "ERROR":
                         stream.stop()
                         # FIXME: Not sure what types we expect here
-                        raise servo.AdjustmentRejectedError(reason=str(deployment))
+                        raise servo.AdjustmentRejectedError(str(deployment))
```
pnickolov, line 1806: doesn't this need a reason code specified?

servo/connectors/kubernetes.py
```diff
@@ -1857,8 +1873,7 @@ def _check_conditions(self, conditions: List[kubernetes_asyncio.client.V1Deploym
                 elif condition.status == "False":
                     raise servo.AdjustmentRejectedError(
                         "ProgressionFailure: message='{condition.status.message}', reason='{condition.status.reason}'",
```
pnickolov, line 1875: should use f-string

servo/connectors/kubernetes.py
```diff
@@ -1901,7 +1916,7 @@ async def raise_for_pod_status(self):
             fmt_str = ", ".join(pod_fmts)
             raise servo.AdjustmentRejectedError(
                 message=f"{len(unschedulable_pods)} pod(s) could not be scheduled for deployment {self.name}: {fmt_str}",
```
pnickolov, line 1918: elsewhere we don't use `message` as a named argument but as positional; I suggest doing it here as well for consistency

pnickolov (changes requested):
Looks good! A new minor notes, primarily on things left over from prior code.

Also, consider moving to enum.Enum for status and reason codes in the future to avoid possible typos (e.g., for words like "unschedulable")

servo/connectors/kubernetes.py
blakewatters: Good catch

servo/connectors/kubernetes.py
blakewatters: We should model the behavior and reraise if its required

servo/connectors/kubernetes.py
blakewatters: Good catch

blakewatters: This also means that we are missing test coverage... we should consider using mocking or something to trigger these code paths if each one of the errors is tough to reproduce as an integration test.

Maybe we factor this off of being an instance method into a standalone function or a new class where we can just unit test it?

servo/connectors/kubernetes.py
linkous8: The rationale for the extra 1 minute buffer in this case is that this particular timeout is applied to all configured optimizations (of which there can be multiple) aggregated under the KubernetesOptimizations class. However each sub-optimization ([DeploymentOptimization](https://github.com/opsani/servox/blob/3c8f47cc9f630e85f0999b2f4709fe47178dda1d/servo/connectors/kubernetes.py#L2453), [CanaryOptimization](https://github.com/opsani/servox/blob/3c8f47cc9f630e85f0999b2f4709fe47178dda1d/servo/connectors/kubernetes.py#L2773)) also has the configured timeout applied to its own operations.

If KubernetesOptimizations' timeout were the exact same as the sub-timeouts, it would be the first to expire and obfuscate information about which sub-optimization was the one that actually caused the timeout

servo/connectors/kubernetes.py
linkous8: Added start-failed as the reason to resolve this for the time being

@blakewatters can you provide more context on what you mean by modelling this out? I'm not sure of what context would even cause this code path to be hit so any additional details would be appreciated

servo/connectors/kubernetes.py
linkous8: Fixed f-string

Opened SOL-356 for Blake's additional note

PR #268, opened 2021-06-15
Bug/fix cancellation

blakewatters:
Restore broken cancellation support. Fixes ENG-9

@linkous8 I guess you are my man for code reviews.

Whenever we moved the errors around into another module we totally missed updating a couple of module paths in the API layer and the exceptions generated were being swallowed by the logging module.

This had the effect of breaking cancellation and the ability to immediately recover from out-of-order commands. I couldn't even remember I had written a lot of this stuff -- it's been broken for months and months no doubt.

It needs some test coverage to bolt this stuff down which is going to involve the FakeAPI backend because we need a sequence of API interactions and are exercising the runner module.

I'll look back to that after I pick off some low-hanging since I've been grinding.


servo/logging.py
```diff
@@ -173,12 +173,16 @@ async def _process_queue(self) -> None:
...
-                logger.warning(f"encountered exception while processing progress logging: {repr(error)}")
+                logger.warning(f"encountered exception while processing progress logging: {repr(progress)} => {repr(error)}")
                 if self._exception_handler:
-                    if asyncio.iscoroutinefunction(self._exception_handler):
-                        await self._exception_handler(error)
-                    else:
-                        self._exception_handler(error)
+                    try:
+                        if asyncio.iscoroutinefunction(self._exception_handler):
+                            await self._exception_handler(progress, error)
```
linkous8, line 180: Need to update the type annotation for the exception_handler argument here: https://github.com/opsani/servox/blob/fdfb83788e858fcf99e5e44f3b05f01a2eba3671/servo/logging.py#L81

servo/runner.py
```diff
@@ -339,19 +345,25 @@ async def _report_progress(**kwargs) -> None:
...
 
+                    # Post a status to resolve the operation
+                    operation = progress['operation']
+                    status = servo.api.Status.from_error(error)
+                    runner = self._runner_for_servo(servo.current_servo())
+                    await runner._post_event(operation, status.dict())
+
                 tasks = [
                     t for t in asyncio.all_tasks() if t is not asyncio.current_task()
                 ]
```
linkous8, line 369: After [cancelling these tasks](https://github.com/opsani/servox/blob/fdfb83788e858fcf99e5e44f3b05f01a2eba3671/servo/runner.py#L371), they should be gathered with return_exceptions set to true to prevent the error message from unretrieved exceptions

linkous8 (changes requested):
Overall the changes look good aside from somewhat minor nitpicks. I'm good with merging once addressed.

PR #278, opened 2021-07-02
Fred/eng 60 servox continuous integration test suite

linkous8:
Temporarily drop poetry caching due to .local install location

As of snok/install-poetry version 1.1.5, the poetry install location is hardcoded to be the entire ~/.local directory instead of being an isolated directory as in previous versions which is problematic for caching: https://github.com/snok/install-poetry/blob/4315377389fc195a3955c4eacf49b86ca8e73f0f/scripts/v1.2/main.sh#L42

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -47,10 +47,10 @@ jobs:
         uses: actions/cache@v2
         id: cached-poetry
         with:
-          path: |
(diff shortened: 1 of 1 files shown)


blakewatters:
We don't want to drop the caching. Fuck all that. Let's solve the problem.

blakewatters:
Superseded by #280

PR #282, opened 2021-07-02
Update resource requirements check to use get config

linkous8:
- Update check to no longer require both request and limit
- Fix test config container name was using image
- Fix result.id assertions
- Validate failure exceptions so they don't pass from random errors
- New tests for resource requirements with get configured
- set_container_resource_defaults_from_config update to use Resource
 enum
- set_container_resource_defaults_from_config fix walrus assignment
 capturing is None boolean instead of existing value
- Update opsani dev connector to check resource requirements using get
 config
-Add integration tests for opsani dev connector resource requirements
 check


blakewatters (changes requested):

blakewatters:
I would do some of the looping logic functionally instead of procedurally if I wrote the code but I ain't mad atcha. Merging

PR #285, opened 2021-07-06
Update logging of errors

linkous8:
- Include reason in error str when specified
- Log status derived from error type before sending
- Include adjust values in raise_for_status for more detailed error msgs
- Include adjust value which caused unschedulable condition in error msg
- raise WatchTimeout in dep rollout so optimization can raise_for_status
- store adjustments applied to optimization on its class
- remove async from adjust abstract method, impelementations are sync
- update insufficient resource tests with more detailed message match


servo/connectors/kubernetes.py
```diff
@@ -2142,7 +2160,7 @@ async def create(
         ...
 
     @abc.abstractmethod
-    async def adjust(
```
blakewatters, line 2145: Why did we drop the async?

servo/connectors/kubernetes.py
```diff
@@ -2463,9 +2482,13 @@ async def apply(self) -> None:
...
-        async with self.deployment.rollout(timeout=self.timeout) as deployment:
-            # Patch the Deployment via the Kubernetes API
-            await deployment.patch()
+        try:
+            async with self.deployment.rollout(timeout=self.timeout) as deployment:
+                # Patch the Deployment via the Kubernetes API
+                await deployment.patch()
+        except WatchTimeoutError:
+            servo.logger.error(f"Timed out waiting for Deployment to become ready...")
+            await self.raise_for_status()
```
blakewatters, line 2491: What's the functional difference here versus the previous raise_for_status in the patch code path? What are we gaining?

servo/connectors/kubernetes.py
```diff
@@ -1892,13 +1898,13 @@ async def raise_for_status(self) -> None:
...
         self._check_conditions(status.conditions)
-        await self.raise_for_pod_status()
+        await self.raise_for_pod_status(adjustments=adjustments)
 
         # Catchall
         self.logger.trace(f"unable to map deployment status to exception. Deployment: {self.obj}")
         raise RuntimeError(f"Unknown Deployment status for '{self.name}': {status}")
 
-    async def raise_for_pod_status(self):
+    async def raise_for_pod_status(self, adjustments: List[servo.Adjustment]):
```
blakewatters, line 1907: I think we may want to rename the method to be more semantically aligned since we're sorta beyond pod status at this point.

Maybe something like `raise_for_failed_pod_adjustments` or something?

servo/connectors/kubernetes.py
```diff
@@ -2129,11 +2138,20 @@ def _normalize_adjustment(adjustment: servo.Adjustment) -> Tuple[str, Union[str,
...
+    Attributes:
+        name (str): The name of the Optimization. Used to set the name for the corresponding component.
+        timeout (Duration): Time interval to wait before considering Kubernetes operations to have failed.
+        adjustments (List[Adjustment]): List of adjustments applied to this optimization (NOTE optimizations are re-created for each
+            event dispatched to the connector. Thus, this value will only be populated during adjust event handling with only the adjustments
+            pertaining to that adjust event dispatch)
     """
 
     name: str
     timeout: servo.Duration
```
blakewatters, line 2152: Drop the whitespace and group the attributes

servo/errors.py
```diff
@@ -72,6 +72,11 @@ def event(self) -> Optional[servo.Event]:
         """The event that was executing when the error occurred."""
         return self._event
 
+    def __str__(self) -> str:
```
blakewatters, line 75: Does it make sense to implement `repr`? Not sure what this looks like when logged

servo/runner.py
```diff
@@ -147,6 +147,7 @@ async def exec_command(self) -> servo.api.Status:
             except servo.errors.EventError as error:
                 self.logger.info(f"Measurement failed: {error}")
                 param = servo.api.Status.from_error(error)
+                self.logger.info(f"Responding with status {param.status}")
```
blakewatters, line 150: I see that we are using the info log level above but does it make more sense to utilize `error`? Not sure what the output looks like

servo/runner.py
```diff
@@ -169,6 +170,7 @@ async def exec_command(self) -> servo.api.Status:
             except servo.EventError as error:
                 self.logger.info(f"Adjustment failed: {error}")
                 status = servo.api.Status.from_error(error)
+                self.logger.info(f"Responding with status {status.status}")
```
blakewatters, line 173: Same comment as above regarding using `error` level vs. `info`

servo/runner.py
```diff
@@ -361,6 +363,7 @@ async def handle_progress_exception(progress: Dict[str, Any], error: Exception)
                     # Post a status to resolve the operation
                     operation = progress['operation']
                     status = servo.api.Status.from_error(error)
+                    self.logger.info(f"Responding with status {status.status}")
```
blakewatters, line 366: Ditto

tests/connectors/kubernetes_test.py
```diff
@@ -1109,7 +1112,10 @@ async def test_adjust_tuning_insufficient_resources(
             setting_name="mem",
             value="128Gi", # impossible right?
         )
-        with pytest.raises(AdjustmentRejectedError, match="Insufficient memory.") as rejection_info:
+        with pytest.raises(
+            AdjustmentRejectedError,
+            match=re.escape("Insufficient memory. Unschedulable adjustment value: fiber-http/fiber-http-tuning.mem=128Gi (Reason unschedulable)")
```
blakewatters, line 1117: Nit but I'd prolly do something like `(Reason: 'unschedulable')` so it easier to parse quickly

blakewatters (changes requested):

servo/connectors/kubernetes.py
linkous8: The implementations and invocations are not async, I've just updated the abstract definition to align with that:

* https://github.com/opsani/servox/blob/bc974b3a10b9a49916d420f705d2c0d35fe3e23d/servo/connectors/kubernetes.py#L2397
* https://github.com/opsani/servox/blob/bc974b3a10b9a49916d420f705d2c0d35fe3e23d/servo/connectors/kubernetes.py#L2560
* https://github.com/opsani/servox/blob/bc974b3a10b9a49916d420f705d2c0d35fe3e23d/servo/connectors/kubernetes.py#L3144

servo/connectors/kubernetes.py
linkous8: The previous raise_for_status in the patch code path does not have access to the adjustments applied to the optimization (nor should it) because it is being called by the Deployment class instance. Adding WatchTimeoutError brings the implementation more in line with that of the tuning optimization's TimeoutError catch block and allows the optimization object to pass the current adjustments into raise_for_status:
https://github.com/opsani/servox/blob/42c20e3dc5a1e466ff81ee971c5f9be34d96b9ec/servo/connectors/kubernetes.py#L2504-L2506

servo/errors.py
linkous8: This change was removed per Peter's ticket feedback; sending reason in the message is redundant and should be logged clearly in the backend

tests/connectors/kubernetes_test.py
linkous8: The reason is no longer included in the message per Peter's ticket feedback; sending reason in the message is redundant and should be logged clearly in the backend

servo/runner.py
linkous8: Logging with error changes the color of the log text from white to red and uses `ERROR` after the timestamp instead of `INFO`. I agree that since the logic is error handling, error is a more appropriate logging method and I've updated the implementation accordingly

servo/runner.py
linkous8: https://github.com/opsani/servox/pull/285#discussion_r668223891

servo/runner.py
linkous8: https://github.com/opsani/servox/pull/285#discussion_r668223891

mjdeclerck (approved):
Approved.  Reviewed with Peter.  Ship it!

PR #288, opened 2021-07-09
Update filterwarnings so test warnings are treated as errors (Fixes ENG-73)

linkous8:


rstarmer (approved):
Looks good to me.  Additional exceptions can always be added if they are found to be needed.

PR #300, opened 2021-07-26
quickstart with even less activation energy

ekalosak:


README.md
```diff
@@ -19,10 +19,16 @@ shared by all servos and a rich library supporting the development of connectors
 
 ## Quick Start
 
+```
```
blakewatters, line 22: add `python` language identifier to fenced code block

README.md
```diff
@@ -19,10 +19,16 @@ shared by all servos and a rich library supporting the development of connectors
...
+pip install servox
```
blakewatters, line 23: I'd just stick with keeping everything oriented around Poetry rather than mixing signals

blakewatters (changes requested):

PR #302, opened 2021-07-28
Relaxed account regex per ENG-83

DanielHHowell:
Modified organization regex for better or worse, along with associated tests


linkous8 (approved):
Would be nice have a comment documenting the concerns we are constraining against but I don't think we should block for that (esp since this is likely to change in the future). Approved

mjdeclerck (approved):
Provisional approval.  Can you please add the validation tests that were performed either in the PR comments or in ENG-83?  I want to make sure that along with the inline documentation in the code is visible for future work/revisions.  Thanks.

linkous8 (approved):
LGTM

PR #303, opened 2021-07-29
Fred/eng 101 argo rollout support

linkous8:
* Update kubernetes_asyncio dependency to custom fork that fixes issue with patching of custom resource
* Add models for parsing rollout custom resource objects from camel case json to snake case to conform with kubernetes python client models
* Add Rollout class to kubernetes connector for working with argo rollouts
* Update CanaryOptimization class to allow it to work with an argo rollout
  * Update Deployment class with new methods introduced for working with rollouts
* Refactor opsani_dev checks to be controller agnostic and add rollout support
* Update cli sidecar injection validation to allow argo rollout as a target
* Add tests and manifests for new rollout support
* Reorder params and rename some existing tests to reduce integration test flakiness caused by kubetest using same namespace/cluster role binding names for multiple tests


servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+# https://github.com/kubernetes-client/python/issues/977#issuecomment-592030030
+class FakeKubeResponse:
+    """Mocks the RESTResponse object as a workaround for kubernetes python api_client deserialization"""
+    def __init__(self, obj):
+        self.data = json.dumps(obj, default=default_kubernetes_json_serializer)
+
+# Use alias generator so that lower camel case can be parsed to snake case properties to match k8s python client behaviour
+def to_lower_camel(string: str) -> str:
+    split = string.split('_')
+    return split[0] + ''.join(word.capitalize() for word in split[1:])
```
blakewatters, line 2037: Prolly should live up in the types module and needs some unit tests. The naming seems unintuitive as it assumes you have a string delimited by an underscore. The comment is perplexing as well. How can you have a `lower camel case` string? Isn't camel-case defined entirely by having a mix of upper and lower case characters delimiting the boundary of the components?

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+        # arbitrary_types_allowed = True
+        alias_generator = to_lower_camel
+        allow_population_by_field_name = True
+
+# Pydantic type models for argo rollout spec: https://argoproj.github.io/argo-rollouts/features/specification/
+# https://github.com/argoproj/argo-rollouts/blob/master/manifests/crds/rollout-crd.yaml
+# NOTE/TODO: fields typed with Any should maintain the same form when dumped as when they are parsed. Should the need
+#   arise to interact with such fields, they will need to have an explicit type defined so the alias_generator is applied
+class RolloutV1LabelSelector(RolloutBaseModel): # must type out k8s models as well to allow parse_obj to work
+    match_expressions: Any
```
blakewatters, line 2050: We can do better than this on the type identifier, no?

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+    requests: Optional[Dict[str, str]]
+
+class RolloutV1Container(RolloutBaseModel):
+    args: Optional[List[str]]
+    command: Optional[List[str]]
+    env: Optional[List[RolloutV1EnvVar]]
+    env_from: Any
+    image: str
+    image_pull_policy: Optional[str]
+    lifecycle: Any
```
blakewatters, line 2094: Why bother to even have these fields if you aren't going to type them? We can just use the `Config` to ignore unspecified fields if you don't want to build them out but specifying fields as `Any` isn't helpful and effectively backslides into turning the data model into a glorified dictionary

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+class RolloutStatusCondition(RolloutBaseModel):
+    last_transition_time: datetime.datetime
+    last_update_time: datetime.datetime
+    message: str
+    reason: str
+    status: str
+    type: str
+
+class RolloutStatus(RolloutBaseModel):
+    HPA_replicas: Optional[int]
```
blakewatters, line 2182: Use the alias functionality to rename the field. We don't want fields with non-idiomatic names

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+    metadata: RolloutV1ObjectMeta
+    spec: RolloutSpec
+    status: Optional[RolloutStatus]
+
+# TODO expose to config if needed
+ROLLOUT_GROUP = "argoproj.io"
+ROLLOUT_VERSION = "v1alpha1"
+ROLLOUT_PURAL = "rollouts"
+
+ROLLOUT_CONST_ARGS = dict(
```
blakewatters, line 2215: Seems like this should be private or constructed where you use it

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+    API Object and provides some state management for the `Rollout`.
+    .. Rollout:
+        https://argoproj.github.io/argo-rollouts/features/specification/
+    """
+
+    obj: RolloutObj
+
+    api_clients: ClassVar[Dict[str, Type]] = {
+        "preferred":kubernetes_asyncio.client.CustomObjectsApi,
+        "argoproj.io/v1alpha1":kubernetes_asyncio.client.CustomObjectsApi,
```
blakewatters, line 2235: Shouldn't you be using the constant here?

servo/connectors/kubernetes.py
```diff
@@ -2524,11 +3109,11 @@ class CanaryOptimization(BaseOptimization):
     """
 
     # The deployment and container stanzas from the configuration
-    deployment_config: "DeploymentConfiguration"
+    deployment_config: Union["DeploymentConfiguration", "RolloutConfiguration"]
     container_config: "ContainerConfiguration"
 
     # State for mainline resources. Read from the cluster
-    deployment: Deployment
+    deployment: Union[Deployment, Rollout]
```
blakewatters, line 3116: This is a bad idea. Add a `rollout` field and then use a validator to ensure that you can configure a `deployment` or a `rollout` but not both. Overloading the `deployment` key like this is no bueno

servo/connectors/kubernetes.py
```diff
@@ -2539,9 +3124,14 @@ class CanaryOptimization(BaseOptimization):
 
     @classmethod
     async def create(
-        cls, deployment_config: "DeploymentConfiguration", **kwargs
+        cls, deployment_config: Union["DeploymentConfiguration", "RolloutConfiguration"], **kwargs
```
blakewatters, line 3127: Rename away from `deployment_config` since it can now be something other than a Deployment.

servo/connectors/kubernetes.py
```diff
@@ -2539,9 +3124,14 @@ class CanaryOptimization(BaseOptimization):
...
     ) -> "CanaryOptimization":
-        deployment = await Deployment.read(deployment_config.name, cast(str, deployment_config.namespace))
+        if isinstance(deployment_config, DeploymentConfiguration):
+            deployment = await Deployment.read(deployment_config.name, cast(str, deployment_config.namespace))
+        elif isinstance(deployment_config, RolloutConfiguration):
+            deployment = await Rollout.read(deployment_config.name, cast(str, deployment_config.namespace))
```
blakewatters, line 3132: Don't overload the name `deployment` like this. Abstract it to something like `target` or even `deployment_or_rollout` so when you read it later you don't falsely assume that it is a Deployment

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+        Return the container with the given name.
+        """
+        return next(filter(lambda c: c.name == name, self.containers), None)
+
+    async def get_tuning_container(self, config: ContainerConfiguration) -> Optional[Container]:
+        tuning_container = self.find_container(config.name)
+        if tuning_container is not None:
+            async with kubernetes_asyncio.client.ApiClient() as api_client:
+                tuning_container.obj = api_client.deserialize(
+                        response=FakeKubeResponse(tuning_container.obj.dict(by_alias=True, exclude_none=True)),
```
blakewatters, line 2407: Wait, what? Why are we returning a fake response? And how is this `get_tuning_container` since it has no direct tie to the concept of a tuning -- which is at the `Pod` not `Container` level?

servo/connectors/kubernetes.py
```diff
@@ -1632,6 +1636,15 @@ def pod_template_spec(self) -> kubernetes_asyncio.client.models.V1PodTemplateSpe
...
 
+    async def get_tuning_pod_template_spec(self) -> kubernetes_asyncio.client.models.V1PodTemplateSpec:
+        """Return a deep copy of the pod template spec for creation of a tuning pod"""
+        return copy.deepcopy(self.pod_template_spec)
+
+    def update_tuning_pod(self, pod: kubernetes_asyncio.client.models.V1Pod) -> kubernetes_asyncio.client.models.V1Pod:
+        """Update the tuning pod with the latest state of the controller if needed"""
+        # NOTE: Deployment currently needs no updating
+        return pod
+
```
blakewatters, line 1647: How are these tied to tuning? It looks like it doesn't know nor care what Pod it is working against and there's nothing bound to tuning at all

servo/connectors/kubernetes.py
```diff
@@ -2550,7 +3140,7 @@ async def create(
         # NOTE: Currently only supporting one container
         assert len(deployment_config.containers) == 1, "CanaryOptimization currently only supports a single container"
         container_config = deployment_config.containers[0]
-        main_container = deployment.find_container(container_config.name)
+        main_container = await deployment.get_tuning_container(container_config)
```
blakewatters, line 3143: `main_container` is assigned to the return value of a method call `get_tuning_container`? Can't be right

servo/connectors/kubernetes.py
```diff
@@ -2566,6 +3156,8 @@ async def create(
             main_container=main_container,
             **kwargs,
         )
+        # Assign config to prevent pydantic Union coercing RolloutConfiguration to DeploymentConfiguration
+        optimization.deployment_config = deployment_config
```
blakewatters, line 3160: Just get rid of the overloaded key and this goes away

servo/connectors/kubernetes.py
```diff
@@ -2698,7 +3290,7 @@ def container_name(self) -> str:
     # TODO: Factor into another class?
     async def _configure_tuning_pod_template_spec(self) -> None:
         # Configure a PodSpecTemplate for the tuning Pod state
-        pod_template_spec: kubernetes_asyncio.client.models.V1PodTemplateSpec = copy.deepcopy(self.deployment.pod_template_spec)
+        pod_template_spec: kubernetes_asyncio.client.models.V1PodTemplateSpec = await self.deployment.get_tuning_pod_template_spec()
```
blakewatters, line 3293: Tuning Pods are a concept that only exist in the canary mode. Pushing that concept up into the `Deployment` model is wrong because it isn't always a thing

servo/connectors/kubernetes.py
```diff
@@ -3087,8 +3682,10 @@ async def create(
         runtime_ids = {}
         pod_tmpl_specs = {}
 
-        for deployment_config in config.deployments:
+        for deployment_config in config.deployments + config.rollouts:
```
blakewatters, line 3685: Don't overload deployment naming

servo/connectors/opsani_dev.py
```diff
@@ -79,32 +91,39 @@ def generate_kubernetes_config(
...
+                    memory=self.memory,
+                )
+            ],
+        )
+        if self.deployment:
+            main_arg = { 'deployments': [ main_config ] }
+        elif self.rollout:
+            main_arg = {
+                'rollouts': [ servo.connectors.kubernetes.RolloutConfiguration.parse_obj(main_config.dict(exclude_none=True)) ],
+                'deployments': []
```
blakewatters, line 118: This should be `None` rather than an empty collection

servo/connectors/opsani_dev.py
```diff
@@ -389,21 +440,21 @@ async def check_kubernetes_service_port(self) -> None:
...
+    async def check_service_routes_traffic_to_controller(self) -> None:
         service = await servo.connectors.kubernetes.Service.read(
             self.config.service, self.config.namespace
         )
-        deployment = await servo.connectors.kubernetes.Deployment.read(
-            self.config.deployment, self.config.namespace
-        )
+        controller = await self.controller_class.read(self.config_controller_name, self.config.namespace)
+        # Remove label dynamically applied by argo-rollouts to service and rollout pods
+        service.selector.pop('rollouts-pod-template-hash', None)
```
blakewatters, line 450: I don't love that Argo Rollouts specific stuff snuck in here when you almost have it fully abstracted. Why is this necessary?

tests/connectors/kubernetes_test.py
```diff
@@ -2013,11 +2015,11 @@ def namespace(self, kube: kubetest.client.TestClient) -> str:
...
     @pytest.mark.parametrize(
-        "service, port",
+        "port, service",
         [
-            ('fiber-http', None),
-            ('fiber-http', 80),
-            ('fiber-http', 'http'),
+            (None, 'fiber-http'),
+            (80, 'fiber-http'),
+            ('http', 'fiber-http'),
```
blakewatters, line 2022: Why'd you flip this? It seems more intuitive to me as service then port?

tests/connectors/kubernetes_test.py
```diff
@@ -2256,3 +2259,146 @@ async def test_telemetry_hello(self, namespace: str, config: KubernetesConfigura
...
+        endpoints = await service.get_endpoints()
+        tuning_name = f"{_rollout_tuning_config.rollouts[0].name}-tuning"
+        tuning_endpoint = next(filter(
+            lambda epa: epa.target_ref.name == tuning_name,
+            endpoints[0].subsets[0].addresses
+        ), None)
+        if tuning_endpoint is None:
+            raise AssertionError(f"Tuning pod {tuning_name} not contained in service endpoints: {endpoints}")
+
+
```
blakewatters, line 2299: Extra whiitespace nit

blakewatters (changes requested):
Comments inline but additionally the code change to test ratio seems out of alignment. There are touches all over the place and a bunch of new classes but it is all backed with a few integration tests. I didn't go and enumerate the cases but I am happy to try and map it out with you.

tests/connectors/kubernetes_test.py
linkous8: This resolves some of the flakiness in integration tests due to kubetest namespace collisions. What happens is that the parameter is appended to the test name then truncated to 43 characters and prepended with kubetest and appended with a timestamp. The result namespace name is `kubetest-test-inject-single-port-deployment-fiber-ht-` which is the same for all 3 parameterizations which causes conflicts when each of the test goes to create the namespace. You can see an example of this failure in this test run: https://github.com/opsani/servox/runs/3204520239

By reordering the parameters, each test parameterization produces a unique namespace name which prevents issues from tests trying to use the same namespace and/or tearing down the namespace set up by another test

pyproject.toml
```diff
@@ -37,6 +36,7 @@ toml = "^0.10.2"
 colorama = "^0.4.4"
 pyfiglet = "^0.8.post1"
 curlify2 = "^1.0.0"
+kubernetes-asyncio = {git = "https://github.com/opsani/kubernetes_asyncio", rev = "v12.1.2-custom-resource-patch-fix"}
```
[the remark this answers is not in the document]
linkous8: The patch I'm using in the fork is not upstreamable as it was applied to the generated code rather than updating the code generation models. The upstream library maintainer has [PRed a fix](https://github.com/kubernetes-client/python/pull/959) for this [issue](https://github.com/tomplus/kubernetes_asyncio/issues/68) into the standard kubernetes python client but its unclear when that will be landed and the fork is mainly a temporary workaround.

The library does not release very frequently so keeping up with patching the fork should not be too burdensome and I've also thrown together some automation on the fork to let us know when we need to rebase it: https://github.com/opsani/kubernetes_asyncio/blob/master/.github/workflows/check-upstream-release.yaml

See [ENG-148](https://linear.app/opsani/issue/ENG-148/kubernetes-connector-is-unable-to-patch-custom-resource-objects-due-to) for discussion pertaining to this workaround

servo/connectors/kubernetes.py
```diff
@@ -2003,6 +2016,578 @@ async def get_restart_count(self) -> int:
...
+    if isinstance(o, (datetime.datetime, datetime.date)):
+        return o.isoformat()
+    raise TypeError(f'Object of type {o.__class__.__name__} '
+                        f'is not JSON serializable')
+
+# https://github.com/kubernetes-client/python/issues/977#issuecomment-592030030
+class FakeKubeResponse:
+    """Mocks the RESTResponse object as a workaround for kubernetes python api_client deserialization"""
+    def __init__(self, obj):
+        self.data = json.dumps(obj, default=default_kubernetes_json_serializer)
```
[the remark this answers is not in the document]
linkous8: This is part of a workaround to allow the use of the `api_client.deserialize()` public method instead of the private method `api_client._ApiClient__deserialize()`. See https://github.com/kubernetes-client/python/issues/977#issuecomment-592030030 for more context and LMK if we should drop the Response faking and just go with the private method instead

servo/connectors/kubernetes.py
linkous8: 1. WRT the method name, the input for this method isn't the field name of the kubernetes object, rather its the fields of the pydantic model. Eg. deletion_grace_period_seconds -> deletionGracePeriodSeconds
2. Lower camel case seemed to be the industry accepted term for having the first word be lower case followed by standard camel case but I will make an update to clarify this with the more formal naming of "dromedary case"

servo/connectors/kubernetes.py
linkous8: RE: the fake response, its also part of the workaround mentioned here: https://github.com/opsani/servox/pull/303#discussion_r685448040

I will update the method name to something more appropriate for the functionality

servo/connectors/kubernetes.py
linkous8: We have no need to interact with these fields in the context of the connector but can't exclude them from the model because they would not be included in copying the rollout template spec into a standalone pod. By leaving the type as Any, we can effectively say we don't care about the typing of these fields because they came from an already validated Kubernetes object and will only ever be used to populate a new Kubernetes object.

I don't see a problem with the model being a glorified dictionary in this context since we are only interested in passing through the data from reads to any write operations and these models are meant to be temporary anyway until the generated models can be implemented.

LMK If you still disagree with the justification above, and I can expand the models to include fields we currently specify as Any.

servo/connectors/kubernetes.py
linkous8: https://github.com/opsani/servox/pull/303#discussion_r685491402

rstarmer (approved):
LGTM

PR #305, opened 2021-07-30
Update prom metrics to support ENG-51 renaming

rstarmer:
All metrics are now averages, except explicit "total" metrics
main_pod_average metric is removed
Requires updates to template to ensure matching metric names
dev2.3 rc live-traffic template being updated to match

### DIFF
--- a/servo/connectors/opsani_dev.py
+++ b/servo/connectors/opsani_dev.py
@@ -133,11 +133,6 @@ class OpsaniDevConfiguration(servo.BaseConfiguration):
                     query='envoy_cluster_membership_healthy{opsani_role="tuning"}',
                 ),
                 servo.connectors.prometheus.PrometheusMetric(
-                    "main_pod_avg_request_rate",
-                    servo.types.Unit.requests_per_second,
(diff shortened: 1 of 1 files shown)


linkous8 (approved):
ENG-51 requirements are addressed, LGTM

DanielHHowell (approved):
Internal consistency achieved

blakewatters:
Hold up -- why would you want to average this? Instead of summing the metrics to produce the true request rate you are now averaging it. The value you are producing is no longer the request rate -- the only way to get that value is to sum up all of the instances to aggregate them. These changes make values that were being reported accurately and semantically aligned with the metric name into one that is not.

If you need averages why not leave the average metrics in place? Why is this in any way necessary and a good idea?

blakewatters (changes requested):
I don't understand this. Explain further but not merging atm

rstarmer:
changes were made to normalize metrics being returned to address ENG-51.
rate for main and tuning need to be of the same order, as that's how
they're used/being compared.  I can change the names, but the intent is the
same.

Robert

PR #307, opened 2021-08-06
Robert/eng 211 dev 23 validation test spec

rstarmer:
Fixes ENG-86, ENG-51, ENG-6, ENG-5

Has a related update in OCO-templates to support the change metrics and removal of p99 metrics


blakewatters:
Tests don't pass and it looks like this is a rollup of other branches

PR #311, opened 2021-08-11
Changes for OKO integration [ENG-166]

snarfbeest:
send control=Control(param['control']) when dispatching 'describe'
event, fix all existing handlers of 'describe' to have **kwargs, so they
can accept this change.

call 'before_event' handlers with the same data as on_event, fix all
existing 'before' handlers to accept this (add **kwargs to the method
signature).

add 'environment' to the types/Control() class (type =
Optional[dict[str, Any]])

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -3715,7 +3715,7 @@ class KubernetesConnector(servo.BaseConnector):
         self.telemetry.remove(f"{self.name}.platform")
 
     @servo.on_event()
-    async def describe(self) -> servo.Description:
+    async def describe(self, **kwargs) -> servo.Description:
         state = await self._create_optimizations()
         return state.to_description()
 
@@ -3725,7 +3725,7 @@ class KubernetesConnector(servo.BaseConnector):
         return state.to_components()
 
     @servo.before_event(servo.Events.measure)
-    async def before_measure(self) -> None:
+    async def before_measure(self, **kwargs) -> None:
         # Build state before a measurement to ensure all necessary setup is done
         # (e.g., Tuning Pod is up and running)
         await self._create_optimizations()
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -1438,6 +1438,16 @@ class Deployment(KubernetesModel):
                 body=options,
             )
 
+    async def scale_to_zero(self) -> None:
+        """this is used instead of 'delete' as the 'destroy' operation to handle 'on_failure: destroy'.
+        Since the Deployment object is used as a wrapper around an existing k8s object that we did not create,
+        it shouldn't be destroyed. Instead, the deployments pods are destroyed by scaling it to 0 replicas.
+        """
+
+        await self.refresh()
+        self.replicas = 0
+        await self.patch()
+
     async def refresh(self) -> None:
         """Refresh the underlying Kubernetes Deployment resource."""
         async with self.api_client() as api_client:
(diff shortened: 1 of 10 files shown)


servo/connectors/kubernetes.py
```diff
@@ -1438,6 +1438,16 @@ async def delete(self, options:kubernetes_asyncio.client.V1DeleteOptions = None)
                 body=options,
             )
 
+    async def scale_to_zero(self) -> None:
+        """this is used instead of 'delete' as the 'destroy' operation to handle 'on_failure: destroy'.
+        Since the Deployment object is used as a wrapper around an existing k8s object that we did not create,
+        it shouldn't be destroyed. Instead, the deployments pods are destroyed by scaling it to 0 replicas.
```
linkous8, line 1444: This needs to be part of a seperate PR with a title/branch name associating it with the relevant issue ([ENG-257](https://linear.app/opsani/issue/ENG-257/servoxkubernetes-modify-on-failure-destroy-behavior-for-deployments))

servo/connectors/kubernetes.py
```diff
@@ -3725,7 +3735,7 @@ async def components(self) -> List[servo.Component]:
         return state.to_components()
 
     @servo.before_event(servo.Events.measure)
-    async def before_measure(self) -> None:
+    async def before_measure(self, **kwargs) -> None:
```
linkous8, line 3738: Why is this being updated when control and environment aren't used by this event handler?

servo/connectors/prometheus.py
```diff
@@ -819,7 +819,7 @@ async def check(
         )
 
     @servo.on_event()
-    def describe(self) -> servo.Description:
+    def describe(self, **kwargs) -> servo.Description:
```
linkous8, line 822: Why is this being updated when control and environment aren't used by this event handler?

servo/types.py
```diff
@@ -1253,6 +1253,10 @@ class files into memory, just-in-time compilation being appliied to critical
     semantics.
     """
 
+    environment: Optional[Dict[str, Any]] = None
```
linkous8, line 1256: Can you model this with a more explicit typing? Environment should be its own model with a "mode" field of type str

linkous8 (changes requested):
The PRed functionality for OKO integration appears to be incomplete. Parameters for the describe and before measure events have been updated but the updated data isn't actually used anywhere. As I said in slack, it would be a lot easier to understand what you're trying to accomplish if I had some documentation of the functionality you're looking to implement

Further, changes for an unrelated issue ([ENG-257](https://linear.app/opsani/issue/ENG-257/servoxkubernetes-modify-on-failure-destroy-behavior-for-deployments)) have crept in, please split these out into their own PR

linkous8:
These changes are insufficient to implement the specified functionality. Please see the associated Linear issue for a list of outstanding items needed for this to be good to merge

servo/connectors/kubernetes.py
snarfbeest: Because (due to a change elsewhere, which is needed to provide the event data in 'before' handlers), this event handler *will* be called with additional arguments - even though it doesn't need them. The way ServoX appears to be designed, ALL event handlers of the same kind (before, on or after) are called with the same exact args - the event dispatcher loops through all of them and calls them with the same exact args (whether the handler wants them or not).

servo/connectors/prometheus.py
snarfbeest: (see reply to above comment)

servo/connectors/kubernetes.py
snarfbeest: It was intended to be "incomplete". We do not want to add this functionality directly into ServoX. FYI: Blake had strong objection to either of the ENG-165 or ENG-166 being integrated into ServoX (and I fully agreed with him on this, too) - they're legacy stuff that is best not to become permanent. On the other hand, the changes here are generic enough and beneficial beyond the specific need for the legacy support - in particular, they allow more to be done in 'before' handlers, which can prove useful in more ways than to simply allow for the functionality described in ENG-165/166 to be implemented.

servo/types.py
snarfbeest: TBH, I would not have modeled 'environment' at all and allowed it (and other extras) to be added and propagated to event handlers - rather than being silently dropped with no warning or notice, as it does now. I added the 'str->any' data type just because it was the easiest change (I know there has to be a way in Pydantic to add 'any extra item allowed', but I haven't bothered finding it).

This may be useful for 'experimental' features (that can be added as connectors, similar to the 'https://github.com/opsani/servo-live-app-emulator' add-on connector) - without requiring a change in the core ServoX code for every tiny little addition to the data.

PR #312, opened 2021-08-13
modify on_failure: destroy behavior for deployments

linkous8:
Since a deployment isn't created by the k8s connector (rather, it
expects to find it already existing on the cluster), the 'destroy'
behavior should not really delete it (among other things, this would
immediately cause an error on the next 'adjust'), but rather destroy the
pods, so that whatever is mis-configured is prevented from running,
while allowing a subsequent 'adjust' to start fresh pods with a different
configuration.

This change modifies on_failure: destroy to scale down the deployment to
zero replicas.

### DIFF
--- a/servo/connectors/kubernetes.py
+++ b/servo/connectors/kubernetes.py
@@ -1438,6 +1438,16 @@ class Deployment(KubernetesModel):
                 body=options,
             )
 
+    async def scale_to_zero(self) -> None:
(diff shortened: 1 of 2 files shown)


snarfbeest:
I opened a new PR for this, against the current head of 'main': #317 (before I saw that Fred opened this one, based on a branch that I had pushed previously). The new one has the same changes, use either one, as preferred (I am typically in favor of the re-based one, to simplify git history).

PR #313, opened 2021-08-16
Changes for OKO integration [preparation for ENG-165, ENG-166]

snarfbeest:
(supersedes closed PR#311)

send control=Control(param['control']) when dispatching 'describe'
event, fix all existing handlers of 'describe' to have a matching arg, so they
can accept this change.

call 'before_event' handlers with the same data as on_event, fix all
existing 'before' handlers to accept this; modify handler type checking to 
compare 'before' handlers to the event's defined 'on' handler type, but with return type=None

add 'environment' to the types/Control() class (type =
Optional[dict[str, Any]])

### DIFF
--- a/servo/events.py
+++ b/servo/events.py
@@ -518,11 +518,14 @@ def event_handler(
         handler_mod_name = handler_localns.get("__module__", None)
         handler_module = sys.modules[handler_mod_name] if handler_mod_name else None
 
-        if preposition == Preposition.before:
-            before_handler_signature = inspect.Signature.from_callable(__before_handler)
+        if preposition in (Preposition.before, Preposition.on):
+            ref_signature=event.signature
+            if preposition == Preposition.before:
+                # 'before' event takes same args as 'on' event, but returns None
+                ref_signature = ref_signature.replace(return_annotation="None")
             servo.utilities.inspect.assert_equal_callable_descriptors(
                 servo.utilities.inspect.CallableDescriptor(
-                    signature=before_handler_signature,
+                    signature=ref_signature,
                     module=event.module,
                     globalns=event_globalns,
                     localns=locals(),
@@ -534,24 +537,7 @@ def event_handler(
                     localns=handler_localns,
                 ),
                 name=name,
-                callable_description="before event handler"
-            )
-        elif preposition == Preposition.on:
-            servo.utilities.inspect.assert_equal_callable_descriptors(
-                servo.utilities.inspect.CallableDescriptor(
-                    signature=event.signature,
-                    module=event.module,
-                    globalns=event_globalns,
-                    localns=locals(),
-                ),
-                servo.utilities.inspect.CallableDescriptor(
-                    signature=handler_signature,
-                    module=handler_module,
-                    globalns=handler_globalns,
-                    localns=handler_localns,
-                ),
-                name=name,
-                callable_description="event handler"
+                callable_description="event handler" if preposition == Preposition.on else "before event handler"
             )
         elif preposition == Preposition.after:
             after_handler_signature = inspect.Signature.from_callable(__after_handler)
@@ -581,10 +567,6 @@ def event_handler(
         return fn
 
     return decorator
-
-
-def __before_handler(self) -> None:
-    pass
 
 
 def __after_handler(self, results: List[EventResult]) -> None:
--- a/servo/events.py
+++ b/servo/events.py
@@ -972,7 +972,7 @@ class _DispatchEvent:
             for connector in self._connectors:
                 try:
                     results = await connector.run_event_handlers(
-                        self.event, Preposition.before
+                        self.event, Preposition.before, *self._args, return_exceptions=False, **self._kwargs
                     )
 
                 except servo.errors.EventCancelledError as error:
--- a/servo/types.py
+++ b/servo/types.py
@@ -1253,6 +1253,10 @@ class Control(BaseModel):
     semantics.
     """
 
+    environment: Optional[Dict[str, Any]] = None
+    """Optional mode control.
+    """
+
     @pydantic.root_validator(pre=True)
     def validate_past_and_delay(cls, values):
         if "past" in values:
--- a/tests/connectors/kubernetes_test.py
+++ b/tests/connectors/kubernetes_test.py
@@ -1570,7 +1570,6 @@ class TestKubernetesConnectorIntegrationUnreadyCmd:
         self,
         tuning_config: KubernetesConfiguration,
         kubetest_deployment_becomes_unready: KubetestDeployment,
-        recwarn: pytest.WarningsRecorder,
         kube: kubetest.client.TestClient
     ) -> None:
         tuning_config.timeout = "25s"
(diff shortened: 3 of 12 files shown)


tests/servo_test.py
```diff
@@ -41,7 +41,7 @@ def adjust_handler(self, results: List[EventResult]) -> None:
         return "adjusting!"
 
     @before_event(Events.measure)
-    def do_something_before_measuring(self) -> None:
+    def do_something_before_measuring(self, metrics: List[str] = [], control: servox.Control = None) -> None:
```
linkous8, line 44: Should use `Control()` instead of None for the sake of consistency

tests/servo_test.py
```diff
@@ -524,7 +524,7 @@ def invalid_measure(self, invalid: str, another: int) -> None:
 
 def test_registering_before_handlers() -> None:
     @before_event("measure")
-    def before_measure(self) -> None:
+    def before_measure(self, metrics: List[str] = [], control: servox.Control = None) -> None:
```
linkous8, line 527: Should use `Control()` instead of None for the sake of consistency

linkous8 (approved):
Aside from some minor nits in servo_test, everything appears to be in order. Approved!

servo/types.py
```diff
@@ -1253,6 +1253,10 @@ class files into memory, just-in-time compilation being appliied to critical
     semantics.
     """
 
+    environment: Optional[Dict[str, Any]] = None
+    """Optional mode control.
+    """
+
```
linkous8, line 1259: Copying relevant bit of feedback from previous PR. snarfbeest wrote:

>TBH, I would not have modeled 'environment' at all and allowed it (and other extras) to be added and propagated to event handlers - rather than being silently dropped with no warning or notice, as it does now. I added the 'str->any' data type just because it was the easiest change (I know there has to be a way in Pydantic to add 'any extra item allowed', but I haven't bothered finding it).
>
>This may be useful for 'experimental' features (that can be added as connectors, similar to the 'https://github.com/opsani/servo-live-app-emulator' add-on connector) - without requiring a change in the core ServoX code for every tiny little addition to the data.

tests/servo_test.py
snarfbeest: The arg is actually unused, which is why I set it to None (unnecessary mem alloc) - but updated it to match other test methods in the same file, anyway. Also dropped "servox." namespace in favor of using the 'from ... import Control' that's already there.

linkous8: Awesome, thanks. This provides greater typing consistency as well since the argument type technically should not be able to accept a None value because its not specified as optional eg. `Optional[Control]`

linkous8:
@pnickolov RE: potential impact to the scripts connector and live-app-emulator, neither connector will be impacted by the updated signature of the before handlers.

The scripts connector uses the same handler signature for all prepositions (before/on/after) which specifies ` *args, **kwargs`. These arguments will pass the handler validation and are not actually used by the connector: https://github.com/opsani/servox/blob/0dde68f3b9851a542697a14c392bf40d1f27551d/servo/connectors/scripts.py#L28

The servo-live-app-emulator connector does not make use of the before preposition in its handlers: https://github.com/opsani/servo-live-app-emulator/blob/main/servo_live_app_emulator/connector.py#L33-L85

blakewatters:
@snarfbeest @pnickolov FYI this change will break other externally maintained connectors such as AppDynamics

servo/events.py
```diff
@@ -518,11 +518,14 @@ def decorator(fn: EventCallable) -> EventCallable:
...
-            before_handler_signature = inspect.Signature.from_callable(__before_handler)
+        if preposition in (Preposition.before, Preposition.on):
+            ref_signature=event.signature
+            if preposition == Preposition.before:
+                # 'before' event takes same args as 'on' event, but returns None
+                ref_signature = ref_signature.replace(return_annotation="None")
             servo.utilities.inspect.assert_equal_callable_descriptors(
                 servo.utilities.inspect.CallableDescriptor(
-                    signature=before_handler_signature,
+                    signature=ref_signature,
```
blakewatters, line 528: What has motivated this change to the before handlers? It has undermined the entire point of the handler chain -- to allow the set of enabled connectors to collaborate through loose coupling.

Adding `Control` as a required input to all event handlers is one thing but now by collapsing the before and on handlers to having the same signature, it is impossible to program before handlers that function abstractly against any event.

This undermines a key part of the design. If you think about use-cases like enforcing a security or timing policy against all operations (e.g., the blackout policy defined by Ancestry), what you want to be able to do is define a single filter callback that applies against all the events. Knowing about each event and defining handlers for each signature breaks this API contract.

What I can get behind is using `*args` and `**kwargs` smartly to allow you to get access to these values if you need them and are programming against a specific event.

tests/connectors/kubernetes_test.py
```diff
@@ -1588,9 +1587,6 @@ async def test_adjust_tuning_settlement_failed(
         with pytest.raises(AdjustmentRejectedError) as rejection_info:
             await connector.adjust([adjustment])
 
-        # Validate raised warnings to ensure all coroutines were awaited
-        assert not any(filter(lambda warn: "was never awaited" in warn.message, recwarn)), list(map(lambda warn: warn.message, recwarn))
-
```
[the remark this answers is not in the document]
linkous8: That error checking logic is now redundant since the pytest configuration was [updated to treat all warnings as errors](https://github.com/opsani/servox/blob/31197324a2d1cb7e1124f0a8f4665bba56d7b396/setup.cfg#L5).

Further, it caused an error during a test run that couldn't be debugged (perhaps due to conflicting logic with the change above) so I advised that the check be removed to let the generic warning handling catch this warning instead: https://github.com/opsani/servox/runs/3352963416

servo/events.py
```diff
@@ -534,24 +537,7 @@ def decorator(fn: EventCallable) -> EventCallable:
...
-                    localns=locals(),
-                ),
-                servo.utilities.inspect.CallableDescriptor(
-                    signature=handler_signature,
-                    module=handler_module,
-                    globalns=handler_globalns,
-                    localns=handler_localns,
-                ),
-                name=name,
-                callable_description="event handler"
```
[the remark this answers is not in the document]
linkous8: These validations only differ by 2 parameters (signature of the first CallableDescriptor, and callable_description). The validation was not removed, it was refactored to reduce the amount of boilerplate code (really all 3 validations could be collapsed into one call to assert_equal_callable_descriptors instead of 3 largely redundant ones).

How does this make validation less robust? Validation for handlers was not deleted...

servo/events.py
linkous8: > it is impossible to program before handlers that function abstractly against any event.

Not really, just add `*args, **kwargs` to the signature of the generic handler and the validation will allow it through. If we're programming against a specific event then we should be able to explicitly type out the parameters expected for that event and have validation in place to ensure those match the event being handled. Using `*args, **kwargs` for specific events seems less robust as the event signature could change which would raise a less informative KeyError during runtime instead of being caught on initialization...

PR #317, opened 2021-08-23
modify on_error:destroy behavior for deployments

snarfbeest:
connectors/kubernetes.py: a deployment handled by the connector is
'foreign' (not created by the connector), the connector cannot
auto-restore it if it has been deleted. Therefore, it should not delete
its deployment-under-control when the on_error setting is configured to
'destroy' (which is actually intended as 'destroy and re-create on the next
adjust').

This patch modifies the connector to set replicas=0, instead -
consistent with the behavior of the servoX predecessor (servo-k8s).

(Linear task: https://linear.app/opsani/issue/ENG-257/servoxkubernetes-modify-on-failure-destroy-behavior-for-deployments)


servo/connectors/kubernetes.py
```diff
@@ -1439,6 +1439,16 @@ async def delete(self, options:kubernetes_asyncio.client.V1DeleteOptions = None)
                 body=options,
             )
 
+    async def scale_to_zero(self) -> None:
+        """this is used instead of 'delete' as the 'destroy' operation to handle 'on_failure: destroy'.
+        Since the Deployment object is used as a wrapper around an existing k8s object that we did not create,
+        it shouldn't be destroyed. Instead, the deployments pods are destroyed by scaling it to 0 replicas.
+        """
```
blakewatters, line 1446: Commenting about the specifics of the `on_failure` semantics here feels like an irrelevant detail about the motivation for introducing the operation is leaking in and violating layering. The `Deployment` model doesn't know or care about `on_failure` -- that is a downstream concept.

Scaling the replicas to zero seems like a perfectly reasonable operation to have as a simple shortcut in the general `Deployment` API though.

One thought to consider on naming off the top of my head is call the operation `shutdown`. I'm not sure if the succinctness is better than the explicit clarity of `scale_to_zero` but throwing it out there as a thought

servo/connectors/kubernetes.py
```diff
@@ -2993,14 +3003,14 @@ async def rollback(self, error: Optional[Exception] = None) -> None:
 
     async def destroy(self, error: Optional[Exception] = None) -> None:
         """
-        Initiates the asynchronous deletion of the Deployment under optimization.
+        Initiates the asynchronous deletion of all pods in the Deployment under optimization.
```
blakewatters, line 3006: This makes sense if I think about it long enough but strikes me as being unintuitive and potentially surprising. When I configure a behavior called `destroy` and it doesn't destroy the target in the configuration context it would seem unintuitive.

What if we invert it by making the `on_failure` behavior for both the Deployment and Canary strategies called `shutdown` or `halt`? In my mind, having an abstract concept of shutting down/halting the strategy in whatever way is contextually appropriate makes deleting a Pod in one case and scaling to zero in another feel more natural.

This is my fault because I reached for symmetry with the `create` action but in hindsight that was a bad call because `destroy` is not a mandatory party of the object lifecycle like `create` is.

`on_failure: shutdown` or `on_failure: halt` both seem clear to me in either canary or deployment contexts where destroy doesn't.

The `Deployment` model operation can still remain `scale_to_zero` to keep that operation explicit. I may have just screwed up the naming of that failure mode enum value and strategy method from the start.

blakewatters (changes requested):

servo/connectors/kubernetes.py
snarfbeest: I realized that part of my proposed changes may cause an issue with compatibility: replacing 'destroy' with 'shutdown' in the configuration definition will cause existing configs that use 'on_failure: destroy' to fail and require them to be updated.
If this could be an issue, we can allow 'destroy' alongside 'shutdown', but mark this as an obsolete feature, to have some time to replace all configurations out there.

servo/connectors/kubernetes.py
```diff
@@ -1439,6 +1439,16 @@ async def delete(self, options:kubernetes_asyncio.client.V1DeleteOptions = None)
                 body=options,
             )
 
+    async def scale_to_zero(self) -> None:
+        """his is used as a "soft" 'delete'/'destroy'.
```
linkous8, line 1443: Typo in this comment. "his" should be "This"

servo/connectors/kubernetes.py
```diff
@@ -3643,13 +3653,17 @@ async def destroy(self, error: Optional[Exception] = None) -> None:
...
+        if self.on_failure == FailureMode.rollback or self.on_failure == FailureMode.shutdown:
             # Ensure that we chain any underlying exceptions that may occur
             try:
                 if self.on_failure == FailureMode.rollback:
                     self.logger.warning(
-                        f"cannot rollback a tuning Pod: falling back to destroy: {error}"
+                        f"cannot rollback a tuning Pod: falling back to shutdown: {error}"
                     )
 
                 await asyncio.wait_for(self.destroy(), timeout=self.timeout.total_seconds())
```
linkous8, line 3669: I think we should call shutdown here for the sake of consistency

linkous8 (approved):
Changes look good overall

servo/connectors/kubernetes.py
linkous8: This is a good point. I agree with leaving support for `destroy` config value but we should log a warning if that value is provided. Then, in a later release, we can fully deprecate it by providing an explicit error message in a custom validator letting users know to update `destroy` to `shutdown`

servo/connectors/kubernetes.py
```diff
@@ -4100,6 +4116,13 @@ class BaseKubernetesConfiguration(servo.BaseConfiguration):
         description="Time interval to wait before considering Kubernetes operations to have failed."
     )
 
+    @pydantic.validator("on_failure")
+    def validate_failure_mode(cls, v):
+        if v == FailureMode.destroy:
+            servo.logger.warning(f"Deprecated value 'destroy' used for 'on_failure', replacing with 'shutdown'")
+            return FailureMode.shutdown
+        return v
```
linkous8, line 4124: Please add a unit test for this. It should probably go here (under TestKubernetesConfiguration):  https://github.com/opsani/servox/blob/0deeb2d5b7af7f0f529e52b28670db84b974f173/tests/connectors/kubernetes_test.py#L396

Something like the following should work:

```
    def test_failure_mode_destroy(self) -> None:
        config = servo.connectors.kubernetes.KubernetesConfiguration(
            namespace="default",
            description="Update the namespace, deployment, etc. to match your Kubernetes cluster",
            on_failure=servo.connectors.kubernetes.FailureMode.destroy,
            deployments=[
                servo.connectors.kubernetes.DeploymentConfiguration(
                    name="fiber-http",
                    replicas=servo.Replicas(
                        min=1,
                        max=2,
                    ),
                    containers=[
                        servo.connectors.kubernetes.ContainerConfiguration(
                            name="fiber-http",
                            cpu=servo.connectors.kubernetes.CPU(
                                min="250m", max="4000m", step="125m"
                            ),
                            memory=servo.connectors.kubernetes.Memory(
                                min="128MiB", max="4.0GiB", step="128MiB"
                            ),
                        )
                    ],
                )
            ],
        )
        assert config.on_failure == FailureMode.shutdown
        assert config.deployments[0].on_failure == FailureMode.shutdown
```

linkous8 (changes requested):
Please add requested unit test. 

Also probably worth resolving the the open items from the last review although I will leave those to your discretion

https://github.com/opsani/servox/pull/317/files#r698604812
https://github.com/opsani/servox/pull/317/files#r698605379

servo/connectors/kubernetes.py
snarfbeest: I thought we were leaving the 'destroy' method named 'destroy', because that's what we do for 'pod shutdown' (per your earlier comment: https://github.com/opsani/servox/pull/317#discussion_r696840440 )

linkous8: Thats correct, the method definition does not need to change but you defined a pod `shutdown` method a few lines above that does not get used that calls the `destroy` method. I'm saying we should use that method in `handle_error` for the sake of consistency with DeploymentOptimization, but the destroy method definition does not need to be changed

https://github.com/opsani/servox/blob/0deeb2d5b7af7f0f529e52b28670db84b974f173/servo/connectors/kubernetes.py#L3656

snarfbeest: Ok, that is an option, too - changing it. (although I defined it merely to ensure that the 'abstract' method is covered by an actual definition, not intending to use the method, actually - since handle_error() does its special thing inline).

linkous8 (approved):
LGTM

PR #332, opened 2021-09-01
daniel/eng-310 EventCancelledError specificity

DanielHHowell:
Changes ServoStatus from generic failed to cancelled (unique from 'aborted' which is utilized in the upcoming fast-fail feature)

### DIFF
--- a/servo/api.py
+++ b/servo/api.py
@@ -31,6 +31,7 @@ class ServoStatuses(str, enum.Enum):
     failed = "failed"
     rejected = "rejected"
     aborted = "aborted"
+    cancelled = "cancelled"
 
 
 Statuses = Union[OptimizerStatuses, ServoStatuses]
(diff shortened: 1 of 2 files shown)


servo/api.py
```diff
@@ -188,7 +191,7 @@ async def report_progress(self, **kwargs) -> None:
             raise servo.errors.UnexpectedEventError(status.reason)
         elif status.status == OptimizerStatuses.cancelled:
             # Optimizer wants to cancel the operation
-            raise servo.errors.EventCancelledError(status.reason)
+            raise servo.errors.EventCancelledError("Command cancelled")
```
linkous8, line 194: Please update this to be `(status.reason or "Command cancelled")` so the backend status is communicated when present

linkous8 (changes requested):
Please expand the unit test here to test EventCancelledError: https://github.com/opsani/servox/blob/9ad1d2f69cd605f24d0a2bacbe60fe1835174311/tests/api_test.py#L7

servo/api.py
DanielHHowell: Good call, thank you

linkous8 (approved):
LGTM

PR #333, opened 2021-09-02
Robert/cm 3 version tag v0.10.7

rstarmer:
tag for release


rstarmer:
Approved as per PR-331 and CM-3

PR #334, opened 2021-09-03
Update if conditions for servox:edge docker build

linkous8:
- Additionally allow the workflow to be run manually

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -11,6 +11,7 @@ on:
   release:
     types: [published]
   pull_request:
+  workflow_dispatch: # allow workflow to be run manually
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -28,7 +29,8 @@ jobs:
   docker:
     needs: pre_job
     if: |
-      needs.pre_job.outputs.should_skip != 'true' &&
+      (needs.pre_job.outputs.should_skip != 'true' ||
```
rstarmer, line 32: just for consistency, may want a space before "needs." and after the parenthesis here.

rstarmer (approved):
Looks good to me, minor non-critical change that might be useful

PR #337, opened 2021-09-08
Add support for Argo Rollouts utilizing workloadRef

linkous8:


DanielHHowell (approved):

servo/connectors/kubernetes.py
```diff
@@ -2484,6 +2550,12 @@ async def inject_sidecar(
             service_port: The port to receive ingress traffic from an upstream service.
         """
 
+        if self.workload_ref_controller:
+            await self.workload_ref_controller.inject_sidecar(
+                name=name, image=image, *args, service=service, port=port, index=index, service_port=service_port
```
rstarmer, line 2555: Does the defintion on line 2529 also need the service_port: Optional[int] = None callout?

tests/connectors/kubernetes_test.py
```diff
@@ -2397,7 +2400,8 @@ async def test_create_rollout_tuning(self, _rollout_tuning_config: KubernetesCon
         if tuning_endpoint is None:
             raise AssertionError(f"Tuning pod {tuning_name} not contained in service endpoints: {endpoints}")
 
-    async def test_adjust_rollout_tuning_cpu_with_settlement(self, _rollout_tuning_config, namespace):
+    async def test_adjust_rol_tuning_cpu_with_settle(self, _rollout_tuning_config, namespace):
+        # test_adjust_rollout_tuning_cpu_with_settlement
```
rstarmer, line 2404: seems this line could probably go if you're cleaning up things.

tests/connectors/opsani_dev_test.py
```diff
@@ -1194,3 +1274,28 @@ async def _remedy_check(id: str, *, config, deployment, kube_port_forward, load_
 
     else:
         raise AssertionError(f"unhandled check: '{id}'")
+
+async def _run_remedy_from_check(failure: servo.Check) -> None:
+    """Replicate the application remedies as done in ServoCLI.check_servo with the remedy argument set to True"""
+    assert failure.remedy, f"Expected check to have remedy method: {failure}"
+
+    if asyncio.iscoroutinefunction(failure.remedy):
+        task = asyncio.create_task(failure.remedy())
```
rstarmer, line 1283: this line is different, which is the right form (may not matter)  with failure.remedy() or failure.remedy

rstarmer (approved):
Few minor comments otherwise, LGTM

servo/connectors/kubernetes.py
linkous8: The `service_port` parameter already has an explicit default value: https://github.com/opsani/servox/blob/95c449049dc4b4041037a3877225c0cdb90fd6ee/servo/connectors/kubernetes.py#L2533

tests/connectors/kubernetes_test.py
linkous8: This line documents the unabbreviated test name in case the abbreviated name causes confusion.

The only reason I have to change the name is to prevent collisions in the kubetest generated namespaces which cause errors unrelated to the servox code during test setup and teardown

tests/connectors/opsani_dev_test.py
linkous8: The right form depends on the state of remedy; whether or not the async function was called before being passed into the check.

If it was not called before/during check instantiation, then it is a coroutine function that would need to be called before being scheduled with create_task. `asyncio.iscoroutinefunction` returns true in this case and remedy is called with `failure.remedy()`.

If remedy was called before/during the check instantiation then it is already coroutine object that simply needs to be scheduled with create_task. In this case, `asyncio.iscoroutinefunction` returns false and `asyncio.iscoroutine` returns true. The remedy is scheduled (`asyncio.create_task(failure.remedy)`) without calling it since it has already been called

rstarmer: Thanks for the clarification!  LGTM

servo/connectors/kubernetes.py
rstarmer: Ah, that works. LGTM

tests/connectors/kubernetes_test.py
rstarmer: That makes sense. LGTM

rstarmer (approved):
Concerns addressed and clarifications make sense. LGTM

PR #345, opened 2021-10-06
Add HPA(KubernetesModel) to servo/connectors/kubernetes.py (ENG-477)

ekalosak:
This PR includes:
- HPA(KubernetesModel) for interacting with HPAs
- time_correction in the servo/repeating.py to include the ability to run cron on a wall clock rather than wait a fixed time between returns.
- README on writing a connector
- adjustment to tests (matching full text of error output is not a good behavior imho, matching the type is good enough)


ekalosak:
ENG-477

servo/repeating.py
```diff
@@ -62,13 +71,28 @@ def start_repeating_task(
...
+                if sleep_time > 0:
+                    await asyncio.sleep(sleep_time)
+                else:
+                    await asyncio.sleep(float_info.epsilon)
 
         asyncio_task = asyncio.create_task(repeating_async_fn(), name=task_name)
         self._repeating_tasks[name] = asyncio_task
         return asyncio_task
 
+
```
linkous8, line 95: Remove this for consistency of whitespacing with other functions in this file

servo/repeating.py
```diff
@@ -62,13 +71,28 @@ def start_repeating_task(
...
+                    await function()
+                elif callable(function):
+                    function()
+                else:
+                    raise TypeError(f"function={function} must be Awaitable or Callable, but has "
+                        f"type(function)={type(function)}.")
+                t1 = time.time()
+                sleep_time = every.total_seconds()
+                if time_correction:
+                    sleep_time -= max(t1 - t0, 0)
```
linkous8, line 85: This should be `sleep_time = max(sleep_time - (t1 - t0), 0)` or drop the `max` since you check for < 0 below and time can only flow forward :wink:

servo/repeating.py
```diff
@@ -62,13 +71,28 @@ def start_repeating_task(
...
+                if sleep_time > 0:
+                    await asyncio.sleep(sleep_time)
+                else:
+                    await asyncio.sleep(float_info.epsilon)
```
linkous8, line 89: Please add a comment indicating this is a concurrency failsafe (assuming that was the reason it was added)

servo/servo.py
```diff
@@ -132,7 +132,6 @@ async def adjust(
     async def promote(self) -> None:
         ...
 
-
```
linkous8, line 135: Sorry for another nitpick but other classes in this file have two whitespace lines between them. Please remove this

tests/servo_test.py
```diff
@@ -539,10 +531,6 @@ def invalid_measure(self, invalid: str, another: int) -> None:
             pass
 
     assert error
-    assert (
-        str(error.value)
-        == """invalid before event handler "before:measure": encountered unexpected parameters "another and invalid" in callable signature "(self, invalid: str, another: int) -> None", expected "(self, *, metrics: 'List[str]' = None, control: 'servo.types.Control' = Control(duration=Duration('0'), delay=Duration('0'), warmup=Duration('0'), settlement=None, load=None, userdata=None, environment=None)) -> 'None'\""""
-    )
```
linkous8, line 545: > adjustment to tests (matching full text of error output is not a good behavior imho, matching the type is good enough)

I'm in full agreement here but all three of these tests raise the same type of error. Can you add some sort of check that the raised error is the one we were expecting? (for all three tests)

linkous8 (changes requested):

servo/connectors/kubernetes.py
```diff
@@ -4248,6 +4323,14 @@ class RolloutConfiguration(BaseKubernetesConfiguration):
...
     replicas: servo.Replicas
 
+class HPAConfiguration(BaseKubernetesConfiguration):
+    """
+    The HPAConfiguration class models the configuration of a modifiable HPA
+    """
+
+    name: str
+    namespace: str
+
```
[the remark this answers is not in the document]
ekalosak: It _is_ used. Kinda the whole point of this PR.

servo/connectors/kubernetes.py
```diff
@@ -4265,11 +4348,14 @@ class KubernetesConfiguration(BaseKubernetesConfiguration):
         description="Argo rollouts to be optimized.",
     )
 
+    hpa: Optional[HPAConfiguration] = pydantic.Field(
+        description="HPA, if present",
+    )
+
```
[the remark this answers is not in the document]
ekalosak: This _is_ used.

servo/repeating.py
```diff
@@ -62,13 +71,28 @@ def start_repeating_task(
...
-                callable()
-                await asyncio.sleep(every.total_seconds())
+                t0 = time.time()
+                if asyncio.iscoroutinefunction(function):
+                    await function()
+                elif callable(function):
+                    function()
+                else:
+                    raise TypeError(f"function={function} must be Awaitable or Callable, but has "
+                        f"type(function)={type(function)}.")
```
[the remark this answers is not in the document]
ekalosak: Coroutines are awaitable: https://docs.python.org/3/library/asyncio-task.html#awaitables

servo/repeating.py
ekalosak: We need a `black` (or other) formatting check in the CI to resolve these kinds of things.

servo/repeating.py
ekalosak: Nope - this reads more clearly as is, the functionality is identical, and the purpose of a review is not to nitpick specific implementation details (especially if they are redundant) but to make sure there aren't errors the tests don't catch.

servo/connectors/kubernetes.py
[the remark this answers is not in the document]
linkous8: Its appears to have been moved to `hpaplus.configuration.HPAConfiguration`

* https://github.com/opsani/connector-hpaplus-pvt/blob/64e5b7df9dbcad0e6896a2f738fae39ade3f52df/hpaplus/checks.py#L99-L101
* https://github.com/opsani/connector-hpaplus-pvt/blob/64e5b7df9dbcad0e6896a2f738fae39ade3f52df/hpaplus/connector.py#L158-L159

servo/connectors/kubernetes.py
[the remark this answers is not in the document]
linkous8: See https://github.com/opsani/servox/pull/345#discussion_r747859588

servo/repeating.py
linkous8: That sounds awesome, LMK what you have in mind

servo/connectors/kubernetes.py
```diff
@@ -766,6 +767,80 @@ def __str__(self) -> str:
...
+        self.logger.info(f'patching HPA "{self.name}"')
+        async with self.api_client() as api_client:
+            api_client.api_client.set_default_header('content-type', 'application/strategic-merge-patch+json')
+            hpa = kubernetes_asyncio.client.V1HorizontalPodAutoscaler()
+            hpa.spec = self.obj.spec
+            hpa_result = await api_client.patch_namespaced_horizontal_pod_autoscaler(
+                name=self.name,
+                namespace=self.namespace,
+                body=hpa,
+            )
```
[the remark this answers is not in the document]
ekalosak: I'll have to test this change, but it's done.

servo/servo.py
ekalosak: We need `black`, this is silly.

tests/servo_test.py
ekalosak: The `with pytest.raises(TypeError) as error` above ensures we're getting the right type of error.

servo/connectors/kubernetes.py
```diff
@@ -766,6 +765,76 @@ def __str__(self) -> str:
...
 
+class HPA(KubernetesModel):
+
+    obj: kubernetes_asyncio.client.V1HorizontalPodAutoscaler
+
+    api_clients: ClassVar[Dict[str, Type]] = {
+        "preferred":kubernetes_asyncio.client.AutoscalingV1Api,
+        "v1":kubernetes_asyncio.client.AutoscalingV1Api,
+        "AutoscalingV2beta1Api":kubernetes_asyncio.client.AutoscalingV2beta1Api,
+        "AutoscalingV2beta2Api":kubernetes_asyncio.client.AutoscalingV2beta2Api,
```
linkous8, line 776: * Please include the full apiVersion in this mapping (eg. "autoscaling/v1")
* Please use the kubernetes apiVersion for this mapping, not the client's python type name
https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields

servo/runner.py
```diff
@@ -341,7 +341,8 @@ def run(self, *, poll: bool = True, interactive: bool = False) -> None:
                 s, lambda s=s: asyncio.create_task(self._shutdown(loop, signal=s))
             )
 
-        loop.set_exception_handler(self._handle_exception)
+        # loop.set_exception_handler(self._handle_exception)
+        loop.set_exception_handler(None)
```
linkous8, line 345: Please revert this or implement it properly with toggle-able config set to use `self._handle_exception` by default

tests/servo_test.py
linkous8: Validating the type of error is not sufficient since the same type `TypeError` is raised by those tests/logic for different reasons. Please add a check that the error was raised for the expected reason so regressions that introduce unexpected `TypeError`s can be caught sooner than later

linkous8 (changes requested):

servo/connectors/kubernetes.py
ekalosak: Keys need to match the API version as would be specified in a `kubectl` command. Using the python kubernetes object names in _not_ correct.

servo/runner.py
ekalosak: Debugging code, nice catch thx

ekalosak:
ok I will add a string check (but not full string match e.g. just a few words n the msg)

servo/connectors/kubernetes.py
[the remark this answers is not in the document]
ekalosak: I'll delete it and everything should still work (modulo some test modifications)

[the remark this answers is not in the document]
ekalosak: Delete the HPA subsection in the KubernetesConfig too

linkous8 (approved):
All good

PR #347, opened 2021-10-08
Fix connector discovery name reflection from class into connector_routes

ekalosak:
Now the `ConncectorLoader` will load the `connector_routes` using keys preferring the `cls.name` before its `__default_name__`
The impact is that if you decorate your new `MySpecialConnector` class with `@servo.metadata(name='myname')` with a name that doesn't match the `cls.__name__.sed('s/Connector//')`, the `servo.yaml` will be allowed to have a connector section under `'myname'` rather than `'myspecial'`.

### DIFF
--- a/servo/cli.py
+++ b/servo/cli.py
@@ -33,8 +33,6 @@ import servo
 import servo.runner
 import servo.utilities.yaml
 
-ENVOY_SIDECAR_IMAGE_TAG = 'opsani/envoy-proxy:servox-v0.9.0'
-
 class Section(str, enum.Enum):
     assembly = "Assembly Commands"
     ops = "Operational Commands"
@@ -1716,6 +1714,9 @@ class ServoCLI(CLI):
             target: str = typer.Argument(
                 ..., help="Deployment or Pod to inject the sidecar on (deployment/NAME or pod/NAME)"
             ),
+            image: str = typer.Option(
+                ..., "--image", "-i", help="Image of the sidecar to be injected in form [REPO]/[IMAGE]:[TAG]"
+            ),
             namespace: str = typer.Option(
                 "default", "--namespace", "-n", help="Namespace of the target"
             ),
@@ -1754,7 +1755,7 @@ class ServoCLI(CLI):
                 )
                 run_async(
                     deployment.inject_sidecar(
-                        'opsani-envoy', ENVOY_SIDECAR_IMAGE_TAG, service=service, port=port
+                        'opsani-envoy', image, service=service, port=port
                     )
                 )
                 typer.echo(f"Envoy sidecar injected to Deployment {deployment.name} in {namespace}")
@@ -1767,7 +1768,7 @@ class ServoCLI(CLI):
                 )
                 run_async(
                     rollout.inject_sidecar(
-                        'opsani-envoy', ENVOY_SIDECAR_IMAGE_TAG, service=service, port=port
+                        'opsani-envoy', image, service=service, port=port
                     )
                 )
                 typer.echo(f"Envoy sidecar injected to Rollout {rollout.name} in {namespace}")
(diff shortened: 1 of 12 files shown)


servo/connector.py
```diff
@@ -386,7 +386,10 @@ def _routes_for_connectors_descriptor(connectors) -> Dict[str, "BaseConnector"]:
             if _validate_class(connector):
                 connector_routes[connector.__default_name__] = connector
             elif connector_class := _connector_class_from_string(connector):
-                connector_routes[connector_class.__default_name__] = connector_class
+                try:
+                    connector_routes[connector_class.name.lower()] = connector_class
+                except AttributeError:
+                    connector_routes[connector_class.__default_name__] = connector_class
```
linkous8, line 392: I'm not sure if this worked for you in testing but there are a few issues with this update:

* This particular routes method `_routes_for_connectors_descriptor` is associated with the configuration of multiple instances of a single connector and is not used in most cases: https://github.com/opsani/servox/#running-multiple-connector-instances
    * In most cases, the routes will come from here https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L489-L496
* Numerous unit tests are currently failing because this update throws off the expected routes of connectors with no metadata.name specified. Eg. [this test connector](https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/tests/servo_test.py#L31) now has a route of firsttestservo instead of first_test_servo
* In most cases, `__default_name__` is set by `_name_for_connector_class` which already uses the connector_class.name property: https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L301-L302 The reason metadata's name is being ignored comes down to order of operations:
    1. In most cases, `__default_name__` is set during `BaseConnector.__init_subclass__`: https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L189
    2. However, metadata is not processed until after `BaseConnector.__init_subclass__` runs so the name used by `_name_for_connector_class` is the default value for name from here: https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L186
    3. Once the metadata decorator sets the name, there is no logic to update the `__default_name__` with the new name from metadata

Given the above points, I think this should be reverted and logic should be added to the [metadata decorator](https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L248-L294) to update `__default_name__` like such:

```
        if name:
            if isinstance(name, tuple):
                if len(name) != 2:
                    raise ValueError(
                        f"Connector names given as tuples must contain exactly 2 elements: full name and alias"
                    )
                cls.name, cls.__default_name__ = name
            else:
                cls.name = name
            cls.__default_name__ = _name_for_connector_class(cls)
```

README.md
```diff
@@ -393,6 +393,27 @@ The bundled connectors are registered and discovered using this mechanism via
...
+```
+in your `pyproject.toml` - you don't need to change the `pyproject.toml` in the servox project.
+This is sufficient for the servo to discover your connector and make it available in `servo.yaml`
+for configuration under the `myspecial:` key. Note that the configuration key is reflected from the
+lexical token representing the class name (e.g. `MySpecialConnector` in this example). To use a
+different connector name, see the following section.
+
+#### Notes on Connector Discovery
+The `connector_routes` loaded by the `servo/connector.py:ConnectorLoader` will use the `echo
+MyConnector | sed 's/Connector//' | lowercase` name for the top-level key in `servo.yaml` unless you
```
linkous8, line 412: This explanation is missing the component which converts the CamelCase names to snake_case here: https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/servo/connector.py#L311

linkous8 (changes requested):
In addition to the line item notes, please add a unit test for the metadata name to the connector unit tests here: https://github.com/opsani/servox/blob/796946547c97e513780b6a6175d2edbd9bd0252d/tests/connector_test.py#L105

blakewatters:
Close it

linkous8:
🕐 💀

PR #357, opened 2021-11-02
Fix success/error rate metrics

linkous8:
- Add "sum by(kubernetes_pod_name)" to success and error rate metrics so
they are averaged by the number of pods instead of averaged by the
number of pods times the number of "envoy_response_code_class"es


rstarmer (approved):
This is great, and was likley much needed!

PR #369, opened 2021-11-30
Kube metrics connector

linkous8:
- Add dateutils as explicit dependency
- Add kube_metrics connector to entry_points
- Implement KubeMetrics connector (alpha)
- Refactor k8s CPU unit model to fulfill new requirements for granularity finer than 1 millicore
  - Improve behavioral consistency of parsing, comparison, and output (always treated as Cores by default instead of conditionally switching between Core and Millicore based on inputs)
- Add __opsani_repr__ to Core and ShortByteSize types
  - Call new __opsani_repr__ from CPU/Memroy __opsani_repr__
- Add validation that adjustments are not finer than 1 millicore in granularity
- Update tests for consistency with updated models


rstarmer (approved):
LGTM

rstarmer (approved):
LGTM.

PR #373, opened 2021-12-07
Danielh/eng 383 generate a tech support object from

DanielHHowell:
Some notes:

- Tried to prevent raising in general on the diagnostics ("assets") API, due to user mis-setting or otherwise as this is kind of a siloed beta functionality and we don't want to bring the whole house down
-  Utilizing [`aiofiles`](https://github.com/Tinche/aiofiles) to maintain proper async
- Corrected the `no_poll` implementation in the `checks` command as `no_diagnostics` functions identically

As I haven't really done servox dev outside of external connectors, likely some ample room for improvement

Can be tested with quick and dirty e-tools script `diagnostics.py`


servo/api.py
```diff
@@ -68,6 +71,15 @@ def response_event(self) -> Events:
...
             raise ValueError(f"unknown command: {self}")
 
+class DiagnosticStates(str, enum.Enum):
+    withhold = "WITHHOLD"
+    send = "SEND"
+    stop = "STOP"
+
+class Diagnostics(pydantic.BaseModel):
+    configmap: Optional[Dict[str, Any]]
+    logs: Optional[Dict[str, Any]]
```
linkous8, line 81: Please move `Diagnostics` to servo/telemetry.py

servo/api.py
```diff
@@ -285,6 +295,104 @@ async def _post_event(self, event: Events, param) -> Union[CommandResponse, Stat
...
+        export_options = dict(exclude_unset=True, exclude_none=True)
+
+        # NOTE: Round-trip through JSON to produce primitives
+        config_dict = self.servo.config.json(**export_options)
+        config_data = json.loads(config_dict)
+
+        return Diagnostics(configmap=config_data, logs=log_dict)
+
+
+    async def _diagnostics_request(self) -> DiagnosticStates:
```
linkous8, line 318: There's a lot of repetition between the methods `_diagnostics_request`, `_post_diagnostics`, and `_reset_diagnostics`. Please refactor the common code into a helper method in servo/telemetry.py

servo/runner.py
```diff
@@ -188,6 +193,35 @@ async def exec_command(self) -> servo.api.Status:
...
+                    asyncio.current_task().cancel()
+                else:
+                    raise
+
+                await asyncio.sleep(10)
+
+            except Exception as error:
+                self.logger.exception(f"failed with unrecoverable error: {error}")
+                raise error
+
```
linkous8, line 224: Please move to a new class (eg. DiagnosticsChecker) in servo/telemetry.py

servo/runner.py
```diff
@@ -234,7 +271,13 @@ def _reraise_if_necessary(task: asyncio.Task) -> None:
         self._main_loop_task = asyncio.create_task(self.main_loop(), name=f"main loop for servo {self.optimizer.id}")
         self._main_loop_task.add_done_callback(_reraise_if_necessary)
 
-    async def run(self, *, poll: bool = True) -> None:
+        if diagnostics:
```
linkous8, line 274: Please move diagnostic task init to the `run` method just above this line https://github.com/opsani/servox/blob/e5d744cb298f6a93d1c55354ba6300f246be67e8/servo/runner.py#L325

servo/runner.py
```diff
@@ -234,7 +271,13 @@ def _reraise_if_necessary(task: asyncio.Task) -> None:
...
+            self._diagnostics_loop_task = asyncio.create_task(self.diagnostics_check(), name=f"diagnostics for servo {self.optimizer.id}")
+            self._diagnostics_loop_task.add_done_callback(_reraise_if_necessary)
```
linkous8, line 276: Diagnostics task should log errors and not re-raise (please add exponential backoff and/or self-termination if there are repeated errors to prevent logs from getting filled with repeat errors)

servo/cli.py
```diff
@@ -1231,6 +1239,23 @@ def check(
...
+                None,
+                "--no-poll",
+                help="Do not poll the Opsani API for commands",
+            ),
+            no_diagnostics: Optional[bool] = typer.Option(
+                None,
+                "--no-diagnostics",
+                help="Do not poll the Opsani API for diagnostics",
+                envvar="NO_DIAGNOSTICS",
+            ),
```
linkous8, line 1252: Please replace this with a new option on `servo.configuration.BaseServoConfiguration`. Analogous shell invocation would be `SERVO_NO_DIAGNOSTICS=True servo run ....`

servo/cli.py
```diff
@@ -1149,6 +1150,12 @@ def run(
                 "--no-poll",
                 help="Do not poll the Opsani API for commands",
             ),
+            no_diagnostics: Optional[bool] = typer.Option(
+                None,
+                "--no-diagnostics",
+                help="Do not poll the Opsani API for diagnostics",
+                envvar="NO_DIAGNOSTICS",
+            ),
```
linkous8, line 1158: Please replace this with a new option on `servo.configuration.BaseServoConfiguration`. Analogous shell invocation would be `SERVO_NO_DIAGNOSTICS=True servo run ....`

linkous8 (changes requested):
In addition to the inline change requests, please move the following to a new class (or classes) in servo/telemetry.py (eg. DiagnosticsChecker). I recommend having it inherit servo.api.Mixin so it has access to the api_client as a member:
* Diagnostics
* _get_diagnostics
* _diagnostics_request
* _post_diagnostics
* _reset_diagnostics
* def running_diagnostics
* async def diagnostics_check

Further, please remove the added `diagnostics` arguments from the various run methods by adding a check on the config here (eg. `self.servo.config.no_diagnostics`): https://github.com/opsani/servox/blob/e5d744cb298f6a93d1c55354ba6300f246be67e8/servo/runner.py#L325

servo/runner.py
DanielHHowell: Ok, I figured it was a part of the API but makes sense to silo. I am having trouble understanding how the Diagnostics class should access the running servo however. If I make a `class DiagnosticsHandler(servo.api.Mixin)` in servo/telemetry.py, what is the correct way for it to operate on the running servo within the runner?

servo/api.py
DanielHHowell: Good call

servo/runner.py
linkous8: You can get the servo for the current context (usually only one context/servo per assembly anyway) by using the `servo.current_servo` method https://github.com/opsani/servox/blob/e5d744cb298f6a93d1c55354ba6300f246be67e8/servo/servo.py#L29

Also note its okay for the runner to continue managing the repeating task (since the runner also [sets up cancellation handling](https://github.com/opsani/servox/blob/e5d744cb298f6a93d1c55354ba6300f246be67e8/servo/runner.py#L393) and needs to be aware of the diagnostics task so its not stopped indefinitely (or at all)). The code run by the task should be part of telemetry.py and the runner is responsible for init (and, optionally, calling a simple start method if it makes sense to implement one outside of the init)

servo/cli.py
DanielHHowell: Ok, I think I've got everything in except this one - likely in need of an implementation hint here

linkous8: Add a new field here: https://github.com/opsani/servox/blob/138f762965073a865b1742f2c9e448d273e2eff3/servo/configuration.py#L491

```
no_diagnostics: bool = pydantic.Field(
    default=False, description="Do not poll the Opsani API for diagnostics"
)
```

It can then be referenced on the configuration of the servo `current_servo.config.no_diagnostics`

(also feel free to invert the name of this if you think it makes more sense `no_diagnostic` -> `diagnostics`)

servo/telemetry.py
```diff
@@ -35,3 +69,163 @@ def remove(self, key: str) -> None:
...
+                elif request == DiagnosticStates.send:
+                    self.logger.info(f"Diagnostics requested, gathering and sending")
+                    diagnostic_data = await self._get_diagnostics()
+
+                    send = await self._put_diagnostics(diagnostic_data)
+                    reset = await self._reset_diagnostics()
+
+                elif request == DiagnosticStates.stop:
+                    self.logger.info(f"Received request to disable polling for diagnostics")
+                    asyncio.current_task().cancel()
```
linkous8, line 109:
```suggestion
                    self.servo.config.no_diagnostics = True
                    self._running = False
```

servo/telemetry.py
```diff
@@ -35,3 +69,163 @@ def remove(self, key: str) -> None:
...
+
+                if request == DiagnosticStates.withhold:
+                    self.logger.trace("Withholding diagnostics")
+
+                elif request == DiagnosticStates.send:
+                    self.logger.info(f"Diagnostics requested, gathering and sending")
+                    diagnostic_data = await self._get_diagnostics()
+
+                    send = await self._put_diagnostics(diagnostic_data)
+                    reset = await self._reset_diagnostics()
```
linkous8, line 105:
```suggestion
                    await self._put_diagnostics(diagnostic_data)
                    await self._reset_diagnostics()
```

servo/telemetry.py
```diff
@@ -35,3 +69,163 @@ def remove(self, key: str) -> None:
...
+
+
+    async def diagnostics_check(self):
+
+        self._running = True
+
+        while self._running:
+            try:
+                self.logger.trace("Polling for diagnostics request")
+
```
linkous8, line 94:
```suggestion
```

tests/api_test.py
```diff
@@ -1,4 +1,7 @@
+import httpx
 import pytest
+import respx
```
linkous8, line 3: Please remove unused imports
```suggestion
import pytest
```

linkous8 (changes requested):

servo/runner.py
```diff
@@ -284,6 +288,12 @@ async def connect() -> None:
         else:
             self.logger.warning(f"Servo runner initialized with polling disabled -- command loop is not running")
 
+        if not servo.current_servo().config.no_diagnostics:
+            diagnostics_handler = servo.telemetry.DiagnosticsHandler(self.servo)
+            self._diagnostics_loop_task = asyncio.create_task(diagnostics_handler.diagnostics_check(), name=f"diagnostics for servo {self.optimizer.id}")
+        else:
+            self.logger.info(f"Servo runner initialized with diagnostics polling disabled")
+
```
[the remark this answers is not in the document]
DanielHHowell: Hmm, it was there originally but moved it to the run based on [first review comment](https://github.com/opsani/servox/pull/373#discussion_r766001691)? What do you think is best?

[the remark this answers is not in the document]
linkous8: You're right, my bad. I think it can be left where it is and instead just remove the task cancellation as part of the main loop startup (or move the cancellation to where the task is being created currently) https://github.com/opsani/servox/blob/ff9ec7b1b305037c3afd984ae124e69b5f6a6ef1/servo/runner.py#L226-L227

servo/runner.py
```diff
@@ -31,7 +32,7 @@ class ServoRunner(pydantic.BaseModel, servo.logging.Mixin, servo.api.Mixin):
     _connected: bool = pydantic.PrivateAttr(False)
     _running: bool = pydantic.PrivateAttr(False)
     _main_loop_task: Optional[asyncio.Task] = pydantic.PrivateAttr(None)
-
+    _diagnostics_loop_task: Optional[asyncio.Task] = pydantic.PrivateAttr(None)
```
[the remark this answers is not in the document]
DanielHHowell: Ah, interesting. I've recently switched to VSCode and the GitLens line comments had me falsely believe one to be already present

servo/runner.py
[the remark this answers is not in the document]
DanielHHowell: I do actually prefer it where it was originally so both loops are on the same tier

servo/telemetry.py
```diff
@@ -1,9 +1,43 @@
...
+import servo.api
+from servo.logging import logs_path
+from servo.types import JSON_FORMAT
+
+
+ONE_MiB = 1048576
+DIAGNOSTICS_MAX_RETRIES = 20
+
+# Intercept backoff decorator logs
+logging.getLogger('diagnostics-backoff').setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
DanielHHowell: The [backoff decorator](https://github.com/litl/backoff#logging-configuration) logs to `INFO` by default and I don't want the logs to crowd with retry attempts for diagnostics at the info level

servo/telemetry.py
```diff
@@ -35,3 +69,163 @@ def remove(self, key: str) -> None:
...
+        for line in log_data_lines:
+            # Handle rare multi-line logs e.g. from self.tuning_container.resources
+            try:
+                time, msg = line.split('|', 1)
+                log_dict[time.strip()] = msg.strip()
+            except:
+                log_dict[list(log_dict.keys())[-1]] += line
+
+        config_dict = self.servo.config.json(exclude_unset=True, exclude_none=True)
+        config_data = json.loads(config_dict)
```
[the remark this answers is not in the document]
DanielHHowell: I tried that, but there were some that did not convert succesfully. In the CLI method Blake had these roundtripped for I imagine the same purpose

servo/telemetry.py
```diff
@@ -35,3 +69,163 @@ def remove(self, key: str) -> None:
...
+
+    @backoff.on_exception(
+        backoff.expo,
+        httpx.HTTPError,
+        max_time=lambda: servo.current_servo() and servo.current_servo().config.settings.backoff.max_time(),
+        max_tries=lambda: DIAGNOSTICS_MAX_RETRIES,
+        logger='diagnostics-backoff',
+        on_giveup=lambda x: asyncio.current_task().cancel()
+    )
+    async def _diagnostics_api(self, client: httpx.AsyncClient, endpoint: str, data: Optional[JSON_FORMAT]=None) -> servo.api.Status:
```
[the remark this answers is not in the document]
DanielHHowell: I marinated on that feedback from the first review, and did extract a helper method for the client.puts() to use, but the response reading and return function differently for `request`, I feel like readability would be reduced.

servo/telemetry.py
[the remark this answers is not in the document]
linkous8: Good thinking. Only issue is you're increasing the verbosity instead of decreasing it: https://loguru.readthedocs.io/en/stable/api/logger.html#levels

I suggest setting to the DEBUG level (also are you sure its getting the same sinks as the main logging assembly so that it gets logged to stdout and the servo.log file?)

servo/telemetry.py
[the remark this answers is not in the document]
linkous8: Use the `request` method with arguments for put and get https://www.python-httpx.org/api/

The error handling is largely the same so just inject the parsing object as a paremeter (the model or a lambda)

```
    async def _diagnostics_request(
        self, 
        method: str, 
        endpoint: str, 
        output_model: pydantic.BaseModel,
        json: Optional[dict] = None,
    ) -> Union[DiagnosticStates, Status]:
        async with self.api_client() as client:
            self.logger.trace(f"{method} diagnostic request")
            try:
                response = await client.request(method=method, url=endpoint, json=json)
                response.raise_for_status()
                response_json = response.json()
                self.logger.trace(
                    f"{method} diagnostic request response ({response.status_code} {response.reason_phrase}): {devtools.pformat(response_json)}"
                )
                self.logger.trace(servo.api._redacted_to_curl(response.request))
                try:
                    return pydantic.parse_obj_as(
                        output_model, response_json
                    )
                except pydantic.ValidationError as error:
                    # Should not raise due to improperly set diagnostic states
                    self.logger.exception(f"Malformed diagnostic {method} response", level_id="DEBUG")
                    return DiagnosticStates.withhold

            except httpx.HTTPError as error:
                self.logger.trace(servo.api._redacted_to_curl(error.request))
                raise
```

servo/telemetry.py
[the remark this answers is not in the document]
linkous8: In that case, I think we can just drop this model conversion alltogether and just add the servo config to the diagnostics object. Then it should be serialized into json by httpx's json parameter

servo/telemetry.py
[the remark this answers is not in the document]
DanielHHowell: Good catch on adding the InterceptHandler. As for the level, they actually say "if you would instead like to log only when a giveup event occurs, set the logger level to ERROR." Tested to verify~

servo/telemetry.py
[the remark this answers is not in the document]
DanielHHowell: I don't like the idea of not pushing it through the exclude filters as one is wading through thick None and Null fields otherwise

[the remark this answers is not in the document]
linkous8: Good point. Further, the model needs to be converted to json by calling pydantic's `.json()` method but httpx will always call `json.dumps()`. Pydantic's json can serialise many commonly used types to JSON (e.g. datetime, date or UUID) which would normally fail json.dumps but theres no way to ensure httpx will use that json encoding implementation (see https://github.com/samuelcolvin/pydantic/issues/951)

Maybe add a todo comment to reevaluate the json round tripping logic later down the road

servo/telemetry.py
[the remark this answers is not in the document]
linkous8: Okay thanks for clarifying, I misunderstood how `setLevel` was working; I thought it was setting the log level of logs emitted by backoff but its actually configuring the minimum severity threshold for backoff's logging filter :+1:

DanielHHowell:
@linkous8 all good to merge here after the integration tests finish?

tests/telemetry_test.py
```diff
@@ -32,3 +32,73 @@ async def test_telemetry_hello(monkeypatch, optimizer: servo.configuration.Optim
...
+        state = servo.telemetry.DiagnosticStates("WITHHOLD")
+        assert state == servo.telemetry.DiagnosticStates.withhold
+
+    def test_send_diagnostic_state(self) -> None:
+        state = servo.telemetry.DiagnosticStates("SEND")
+        assert state == servo.telemetry.DiagnosticStates.send
+
+    def test_stop_diagnostic_state(self) -> None:
+        state = servo.telemetry.DiagnosticStates("STOP")
+        assert state == servo.telemetry.DiagnosticStates.stop
```
linkous8, line 47:
```suggestion
@pytest.mark.parametrize(
    "input, expected_state",
    [
        ("WITHHOLD", servo.telemetry.DiagnosticStates.withhold),
        ("SEND", servo.telemetry.DiagnosticStates.send),
        ("STOP", servo.telemetry.DiagnosticStates.stop),
    ],
)
def test_withhold_diagnostic_state(input, state) -> None:
    state = servo.telemetry.DiagnosticStates(input)
    assert state == expected_state
```

tests/telemetry_test.py
```diff
@@ -32,3 +32,73 @@ async def test_telemetry_hello(monkeypatch, optimizer: servo.configuration.Optim
...
+async def test_diagnostics_request(monkeypatch, optimizer: servo.configuration.Optimizer) -> None:
+
+    # Simulate running as a k8s pod
+    monkeypatch.setenv("POD_NAMESPACE", "test-namespace")
+
+    # NOTE: Can't use servo_runner fixture; its init happens prior to setting of env var above
+    servo_runner = servo.runner.ServoRunner(servo.Servo(
+        config=servo.BaseServoConfiguration(name="archibald", optimizer=optimizer),
+        connectors=[], # Init empty servo
+    ))
```
linkous8, line 59: Don't need to simulate running in k8s for these tests

```suggestion
async def test_diagnostics_request(servo_runner: servo.runner.ServoRunner, optimizer: servo.configuration.Optimizer) -> None:
```

linkous8 (approved):
It would be nice to have a test that runs the `diagnostics_check` for a quick end to end verification but I'm not going to block on that. Feel free to implement post 2.5

(Also it may be easier to test that method if you move the while loop to another method to make it easier to run a single check eg. `async def diagnostics_check_loop(self)`  and `async def diagnostics_check(self)`)

https://github.com/opsani/servox/blob/df5f3bebd733e9a5ee296bc59c03b0ade076877d/servo/telemetry.py#L89

PR #377, opened 2021-12-14
ENG-700

ekalosak:
- made servo issue a startup event when a run gets cancelled;
- made the PrometheusConnector.startup() idempotent


servo/runner.py
```diff
@@ -223,6 +224,8 @@ async def main_loop(self) -> None:
     def run_main_loop(self) -> None:
         if self._main_loop_task:
             self._main_loop_task.cancel()
+            loop = asyncio.get_event_loop()
+            loop.create_task(self.servo.dispatch_event(servo.Events.startup))
```
linkous8, line 228:
```suggestion
    async def run_main_loop(self) -> None:
        if self._main_loop_task:
            self._main_loop_task.cancel()
            
        await self.servo.startup()
        self.logger.info(
            f"Servo started with {len(self.servo.connectors)} active connectors [{self.optimizer.id} @ {self.optimizer.url or self.optimizer.base_url}]"
        )
```

I agree we should update the startup/shutdown lifecycle given the implementation of cancel responses. However, we need to update a few more places for the sake of completeness:

* We're moving this startup into `run_main_loop()` (delete the old one): https://github.com/opsani/servox/blob/6ce0cbc39fd003290d0f54a9dcafb19278f399da/servo/runner.py#L257
* await run_main_loop now that its async:
  * https://github.com/opsani/servox/blob/6ce0cbc39fd003290d0f54a9dcafb19278f399da/servo/runner.py#L299
  * https://github.com/opsani/servox/blob/6ce0cbc39fd003290d0f54a9dcafb19278f399da/servo/runner.py#L404
* shutdown the servo during cancellation handling: 
  * https://github.com/opsani/servox/blob/6ce0cbc39fd003290d0f54a9dcafb19278f399da/servo/runner.py#L373-L404
  * https://github.com/opsani/servox/blob/6ce0cbc39fd003290d0f54a9dcafb19278f399da/servo/runner.py#L518-L527

linkous8 (changes requested):

blakewatters:
What are we doing with this? Seems pretty reasonable...

Except I don't love that each connector has to be aware of being potentially restarted and guard against it. Seems like a lifecycle teardown of the channel would keep it more straightforward or even a blanket teardown of all channels.

It shouldn't be the connector's responsibility to handle this state. It would have to be replicated everywhere

linkous8:
Looking into the relevant internal project mgmt, it would seem @ekalosak reached a similar conclusion in that this should be handled externally. I'm stilling looking into it to see if it has impact on new startup behavior for the k8s connector

PR #394, opened 2022-01-18
Enable container log for unready integration tests

linkous8:


DanielHHowell (approved):
LGTM. What is the cause of the connection error in the xfail however?

linkous8:
Intermittent issue with the minikube environment that occurs before any test code is even run. I've been trying to squish all these flukes so rerunning tests doesn't become a regular part of the PR process

PR #400, opened 2022-01-25
Log payload of unretryable status codes

linkous8:
- Add missing servo import
- Set unretryable log level to error
- Add error response to log
- Add .vscode to project for auto linting with black and other useful
settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 node_modules
 .DS_Store
-.vscode
+.vscode/launch.json
 __pycache__
 .coverage
 /servo.yaml
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,12 +1,12 @@
 {
     "workbench.colorCustomizations": {
-        // For ext stuart.unique-window-colors, can be removed
+        "_comment": "For ext stuart.unique-window-colors, can be removed",
         "activityBar.background": "#24320E",
         "titleBar.activeBackground": "#324613",
         "titleBar.activeForeground": "#F7FBF1"
     },
     "python.analysis.diagnosticSeverityOverrides": {
-        "reportUnusedCoroutine": "information",
+        "reportUnusedCoroutine": "information"
     },
     "python.formatting.provider": "black",
     "editor.formatOnSave": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,16 @@
+{
+    "workbench.colorCustomizations": {
+        // For ext stuart.unique-window-colors, can be removed
+        "activityBar.background": "#24320E",
+        "titleBar.activeBackground": "#324613",
+        "titleBar.activeForeground": "#F7FBF1"
+    },
+    "python.analysis.diagnosticSeverityOverrides": {
+        "reportUnusedCoroutine": "information",
+    },
+    "python.formatting.provider": "black",
+    "editor.formatOnSave": true,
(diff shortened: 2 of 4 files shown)


pavel-georgiev (commented):
What is the output when there is a >500 error? 

Also - is this safe for the case when the error has no payload?

linkous8:
> What is the output when there is a >500 error?

The response object is appended to the current log message as a pretty formatted string: https://docs.python.org/3/library/pprint.html#pprint.pformat


> Also - is this safe for the case when the error has no payload?

There is no increased risk over the current implementation. If you have a clearly documented safety concern then please raise a new issue for triage

pavel-georgiev (changes requested):
Fred, this has the following issues that are not as specified in ENG-790:
1. The error payload for < 500 codes in not logged. Instead you get a dump of the `requests` response object, i.e. `<Response [400 Bad Request]>`, which is not helpful. The intent of this change is to log the payload. Also (I tried to hint that in my original note) - when logging the response body, make sure it can work with an empty payload
2. There is no response logging at all for errors >= 500 

As a general note - please test these changes before submitting a fix for review. If you need a way to simulate OCO responding with an error, see the original issue description (ENG-754), the issue is easily reproducible. If you want to have OCO return > 500 instead of 400 (for testing purposes), please see [this](https://github.com/opsani/oco/commit/4eef26fd8af2b2b945c8572361f4c433008b132b) OCO branch

linkous8:
> 1. The error payload for < 500 codes in not logged. Instead you get a dump of the `requests` response object, i.e. `<Response [400 Bad Request]>`, which is not helpful. The intent of this change is to log the payload. Also (I tried to hint that in my original note) - when logging the response body, make sure it can work with an empty payload

My bad, NMF these web libraries make it so difficult to dump the entire response content. You'd think devtools would work properly here...

> 2. There is no response logging at all for errors >= 500

* https://github.com/opsani/servox/blob/6901d5f76cb886a8f11c5682c6401e139c3303a9/servo/api.py#L283-L304
* https://github.com/opsani/servox/blob/896266cd2f68c13cb036b099f7806c8a44e9b6c3/servo/api.py#L267

linkous8:
> As a general note - please test these changes before submitting a fix for review. If you need a way to simulate OCO responding with an error, see the original issue description (ENG-754), the issue is easily reproducible. If you want to have OCO return > 500 instead of 400 (for testing purposes), please see [this](https://github.com/opsani/oco/commit/4eef26fd8af2b2b945c8572361f4c433008b132b) OCO branch

As a general note, please attach FULL TRACE LEVEL logs when you expect someone to go fix something for you

pavel-georgiev (changes requested):
@linkous8 As is, Servo still does not appear to log payload for >= 500 errors (as requested by ENG-790).

To avoid more cycles on this, please fix this and post sample output of how servo handles the errors (please test with 4xx and 5xx errors).

DanielHHowell:
At risk of being too many cooks in the kitchen -- just trying to mediate here -- all that should be required to get this to the ticket specification is moving the payload logging from the giveup function to the main error block so it's caught no matter the status code.

linkous8:
@DanielHHowell LGTM thank you :pray: 

@pavel-georgiev does this capture your completely unwritten requirements or will you require another half hour meeting to verbally communicate your concerns?

pavel-georgiev (approved):

PR #402, opened 2022-02-08
Danielh/eng 785 run mypy on servox

DanielHHowell:


.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,121 @@
...
+        id: cached-project
+        with:
+          path: .venv
+          key: ".venv-${{ runner.os }}-\
+            python-${{ steps.setup-python.outputs.python-version }}-\
+            ${{ hashFiles('**/poetry.lock', 'servo/**/*.py') }}"
+      - name: Install root project
+        # NOTE: Skipping root project install accelerates cases where only tests have changed
+        if: steps.cached-project.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction
```
linkous8, line 88: I know I initially requested mypy have its own file but didn't realize the dependencies would need to be fully installed for type analysis. Please move the mypy job to the test.yaml workflow file to avoid defining and running the prejob and setup build steps redundantly

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,121 @@
...
+      with:
+        path: .venv
+        key: ".venv-${{ runner.os }}-\
+          python-${{ steps.setup-python.outputs.python-version }}-\
+          ${{ hashFiles('**/poetry.lock', 'servo/**/*.py') }}"
+    - name: Run mypy
+      run: |
+        source .venv/bin/activate
+        # NOTE: We just need to get a report from mypy (not fail at the plethora), so success is forced
+        mypy servo/ || true
```
linkous8, line 117: Please have this fail (or do some kind of yellow warning if GH actions supports it). It won't be a required test for merge for some time

linkous8 (changes requested):
Overall LTGM to me, I doubt we will get the tests green until after the upcoming architectural refactor (so will leave them optional for now)

linkous8 (approved):

PR #411, opened 2022-03-16
Danielh/eng 833 coalesce multiple opsani dev remedies to

DanielHHowell:


linkous8 (changes requested):
from a cursory glance a couple of items:
* I dont see any automated unit/integration tests added that cover both types of the configuration
* You added a CLI flag for the checks which is good but I thought there was some kind of check configuration section for the servo.yaml. Can we get it added there as well as it may be required for environment variable configuration

DanielHHowell:
> from a cursory glance a couple of items:
> 
> * I dont see any automated unit/integration tests added that cover both types of the configuration
> * You added a CLI flag for the checks which is good but I thought there was some kind of check configuration section for the servo.yaml. Can we get it added there as well as it may be required for environment variable configuration

@linkous8 I haven't found a check configuration outside of the CLI flag. Also I don't see any tests that cover the flags of the relevant CLI section, will see about adding that. Do you think an integration test running the two variants of checking on say, fiber-http would be good?

linkous8:
> @linkous8 I haven't found a check configuration outside of the CLI flag.

Lets go ahead and add a new ChecksConfiguration model for that, say, around here: https://github.com/opsani/servox/blob/d027184c45e0bd2e15724a7a4c9af336602cf57a/servo/configuration.py#L467

Long term, I'm looking to streamline the CLI entrypoint to make the flow more understandable and will be preferring file/env var based configuration over CLI flags as the latter is more difficult to keep track of and alter programmatically

> Also I don't see any tests that cover the flags of the relevant CLI section, will see about adding that. Do you think an integration test running the two variants of checking on say, fiber-http would be good?

See if you can adapt this test to test both variants of the configuration: https://github.com/opsani/servox/blob/d027184c45e0bd2e15724a7a4c9af336602cf57a/tests/connectors/opsani_dev_test.py#L1176

servo/checks.py
```diff
@@ -735,6 +737,152 @@ class Config:
...
+    @classmethod
+    async def process_checks(
+        cls,
+        checks_config: servo.configuration.ChecksConfiguration,
+        results: list,
+        passing: set,
+    ) -> bool:
+
+        remedy = checks_config.remedy
+        check_halting = checks_config.check_halting
```
linkous8, line 750: These variables are used once each, please reference the config directly for brevity and clarity on lines 791 and 799

servo/checks.py
```diff
@@ -735,6 +737,152 @@ class Config:
         extra = pydantic.Extra.allow
 
 
+class CheckHelpers(pydantic.BaseModel, servo.logging.Mixin):
+    @classmethod
+    async def process_checks(
+        cls,
+        checks_config: servo.configuration.ChecksConfiguration,
+        results: list,
+        passing: set,
```
linkous8, line 746:
```suggestion
        results: list[servo.events.EventResult],
        passing: set[str],
```

servo/checks.py
```diff
@@ -31,6 +32,7 @@
 )
 
 import pydantic
+from tabulate import tabulate
 
 import servo.configuration
 import servo.logging
```
linkous8, line 38: add import for servo.events (note the event reduction may be better located in the servo if events import turns into a circular reference anywhere)

servo/checks.py
```diff
@@ -735,6 +737,152 @@ class Config:
...
+
+        remedy = checks_config.remedy
+        check_halting = checks_config.check_halting
+
+        ready = False
+        failure = None
+
+        checks: List[servo.Check] = functools.reduce(
+            lambda a, b: a + b.value, results, []
+        )
```
linkous8, line 757:
```suggestion
        checks: list[Check] = functools.reduce(lambda a, b: a + b.value, results, [])
```

servo/checks.py
```diff
@@ -735,6 +737,152 @@ class Config:
...
+                    break
+
+        if not failure:
+            servo.logger.info("🔥 All checks passed.")
+            ready = True
+
+        return ready
+
+    @classmethod
+    async def checks_to_table(cls, checks_config, results) -> Tuple[bool, str]:
```
linkous8, line 809:
```suggestion
    async def checks_to_table(cls, checks_config, results) -> str:
```

servo/checks.py
```diff
@@ -735,6 +737,152 @@ class Config:
...
+                message = functools.reduce(
+                    lambda m, e: m + f"({errors.index(e) + 1}/{len(errors)}) {e}\n",
+                    errors,
+                    "",
+                )
+                row = [result.connector.name, status, message]
+                table.append(row)
+
+        # Output table
+        if not quiet:
```
linkous8, line 879: Move this check to line 476 of servo.py. to_table should always return a table

linkous8 (approved):
A few minor nitpicks aside, LGTM! 🚀

servo/servo.py
```diff
@@ -403,6 +402,102 @@ def top_level_schema_json(self, *, all: bool = False) -> str:
...
+                checks_config=self.config.checks,
+                results=results,
+                passing=passing,
+            )
+            if not progressive:
+                output = await servo.checks.CheckHelpers.checks_to_table(
+                    checks_config=self.config.checks, results=results
+                )
+                if not quiet:
+                    print_callback(output)
```
linkous8, line 482:
```suggestion
            if not progressive and not quiet:
                output = await servo.checks.CheckHelpers.checks_to_table(
                    checks_config=self.config.checks, results=results
                )
                print_callback(output)
```

linkous8: No need to process table if quiet

PR #412, opened 2022-03-17
Add "skipped" conditional for near zero metric value fast fails

DanielHHowell:


servo/fast_fail.py
```diff
@@ -15,11 +15,13 @@
...
 
 SLO_FAILED_REASON = "slo-violation"
+SLO_MINIMUM_VALUE = 0.25
 
 
 class SloOutcomeStatus(str, enum.Enum):
     passed = "passed"
     failed = "failed"
+    skipped = "skipped"
     missing_metric = "missing_metric"
```
linkous8, line 25:
```suggestion
    zero_metric = "zero_metric"
    zero_threshold = "zero_threshold"
    missing_metric = "missing_metric"
```

servo/fast_fail.py
```diff
@@ -104,6 +117,32 @@ def check_readings(
...
+                    self._results[condition].append(
+                        SloOutcome(
+                            **result_args, status=SloOutcomeStatus.missing_threshold
+                        )
+                    )
+                    continue
+
+                elif (float(threshold_value) == 0) and (
+                    0 < metric_value <= condition.slo_metric_minimum
+                ):
```
linkous8, line 130: Could you elaborate on the logic behind these checks? Seems to me like you only need the second part of the `and` and should be checking threshold_value instead of metric_value but maybe I'm missing something?

servo/types/slo.py
```diff
@@ -17,12 +17,14 @@ class SloKeep(str, enum.Enum):
...
     description: Optional[str] = None
     metric: str
+    slo_metric_minimum: float = 0.25
     threshold_multiplier: decimal.Decimal = decimal.Decimal(1)
     keep: SloKeep = SloKeep.below
     trigger_count: pydantic.conint(ge=1, multiple_of=1) = 1
     trigger_window: pydantic.conint(ge=1, multiple_of=1) = None
     threshold: Optional[decimal.Decimal]
     threshold_metric: Optional[str]
+    slo_threshold_minimum: Optional[float] = 0.25
```
linkous8, line 27: Don't need Optional if we're setting a default. Value should never be null
```suggestion
    slo_threshold_minimum: float = 0.25
```

servo/types/slo.py
```diff
@@ -42,6 +44,19 @@ def _check_threshold_values(cls, values):
...
+    @pydantic.root_validator
+    @classmethod
+    def _check_duplicated_minimum(cls, values):
+        if (
+            values.get("threshold") is not None
+            and values.get("slo_threshold_minimum") is not None
+        ):
+            raise ValueError(
+                "SLO Condition cannot specify both static threshold and metric based threshold minimum"
+            )
```
linkous8, line 56: Also try and include some sort of identifier for the condition in the warning message
```suggestion
            # Use run time import to prevent circular imports
            import servo.logging
            servo.logging.logger.warning(
                "SLO Condition should not specify both static threshold and metric based threshold minimum."
                " Please double check the Slo Conditions of the User Config"
            )
```

linkous8 (changes requested):

servo/fast_fail.py
```diff
@@ -116,7 +120,13 @@ def check_readings(
 
             # Check target against threshold
             check_passed_op = _get_keep_operator(condition.keep)
-            if check_passed_op(metric_value, threshold_value):
+            if (float(threshold_value) == 0) and (
```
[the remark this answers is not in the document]
linkous8: > (also add a config pydantic condition validation to warn if the static threshold and threshold minimum are both set simultaneously as that is a likely configuration error)

See https://github.com/opsani/servox/pull/412#discussion_r840688682

servo/fast_fail.py
```diff
@@ -104,9 +121,32 @@ def check_readings(
...
+                    )
+                    continue
+
+                elif 0 < metric_value <= condition.slo_metric_minimum:
+                    self._results[condition].append(
+                        SloOutcome(**result_args, status=SloOutcomeStatus.zero_metric)
+                    )
+                    continue
+
+                elif 0 < threshold_value <= condition.slo_threshold_minimum:
```
linkous8, line 143:
```suggestion
                elif 0 <= metric_value <= condition.slo_metric_minimum:
                    self._results[condition].append(
                        SloOutcome(**result_args, status=SloOutcomeStatus.zero_metric)
                    )
                    continue

                elif 0 <= threshold_value <= condition.slo_threshold_minimum:
```

linkous8 (approved):

PR #424, opened 2022-04-26
First pass, autosetting within canaryopt 'create' classmethod

DanielHHowell:


DanielHHowell:
@linkous8 can you let me know if this is the right idea, mainly as far as where this logic should take place? And for the actual multipliers (just hardcoded temporarily for testing), do you think they should be constants somewhere or perhaps part of the config?

DanielHHowell:
@linkous8 I moved the logic to `startup`. Any advice on how/where we should store the autosetting multipliers (constants, config, etc)? I also added a small touch to just load default ranges when `create_tuning_pod` is False, doesn't really matter since the values aren't used anyways, just to hold the important pinned settings

servo/connectors/opsani_dev.py
```diff
@@ -116,18 +116,28 @@ def generate_kubernetes_config(
...
 
         else:
             # NOTE: currently assuming we NEVER want to adjust the main deployment with the opsani_dev connector
             # TODO: Do we ever need to support opsani dev bootstrapping of non-canary adjusted optimization of deployments?
+
+            # Just load defaults when create_tuning_pod is False - these values aren't used and just hold the pinned setting
+            if not self.cpu:
+                self.cpu = CPU(min="250m", max="3000m")
+            if not self.memory:
+                self.memory = Memory(min="256 MiB", max="3.0 GiB")
```
blakewatters, line 139: why have these famous random values versus just raising?

blakewatters:
`startup` seems correct here. I'd prolly argue for putting the multipliers into the config with defaults expecting that you never have to change them but you can if you have to. If you hardcode them with constants you need to cut new bits to experiment

PR #425, opened 2022-04-27
Add coverage report + artifact upload to integration and system tests

DanielHHowell:


linkous8 (changes requested):
@DanielHHowell thanks so much for adding this! Results look good so far!

image


However there is a bit of room for improvement https://stackoverflow.com/a/58278633

DanielHHowell:
@linkous8 that coverage appending looks like a great feature! But it seems this would require having `pytest` run in the same job/env as the unit tests -- do we want to combine those then?

linkous8:
> @linkous8 that coverage appending looks like a great feature! But it seems this would require having `pytest` run in the same job/env as the unit tests -- do we want to combine those then?

A very good point. Github provides the artifacts as a means of transferring the environment details between workflow jobs. You should be able to download the report from the unit test artifact to be appended to by the integration tests (note this will likely require we make the integration tests dependent upon the completion of the unit test before they start their run)

linkous8:
Also meant to say it should fall back to running standalone if the unit test job fails

.github/workflows/test.yaml
```diff
@@ -212,22 +220,91 @@ jobs:
...
         echo "$KUBE_CONFIG_DATA" | base64 --decode > $HOME/.kube/config
+        sed -i -e 's/v1alpha1/v1beta1/' $HOME/.kube/config
         kubectl config view --minify --flatten > tests/kubeconfig
       shell: bash
       env:
         KUBE_CONFIG_DATA: ${{ secrets.KUBE_CONFIG_DATA }}
-    - name: Install AWS IAM Authenticator
-      uses: prepor/action-aws-iam-authenticator@master
+    - name: Install aws-iam-authenticator
+      # Note: pin to v0.5.7
```
blakewatters, line 229: put it into a var you can reference in one place

pyproject.toml
```diff
@@ -38,6 +38,7 @@ curlify2 = "^1.0.0"
 kubernetes_asyncio = "^12.1.2"
 aiofiles = "^0.8.0"
 python-dateutil = "^2.8.2"
+coverage = "^6.3.3"
```
blakewatters, line 41: This should be a development dependency not a primary

blakewatters:
I would strongly advise against making the integration tests dependent upon completion of the units. I made this test suite wicked fast by parallelizing the shit out of it.

What you want to do (and I don't know how to get there but pretty sure its possible) is to add a tail job that waits for the unit/integration reports to be published and then kicks in to combine and report them once they are available.

Serializing the suite is trading off speed that you need day to day for a coverage report that let's be honest you prolly aren't gonna do shit about.

Don't cripple the workflow. Always defend the territory you have already taken and only trade it off explicitly when the upside is just too good to pass off.

PR #426, opened 2022-05-03
Set requests=limits if unspecified, do not return metrics if limits u…

DanielHHowell:
…nspecified

### DIFF
--- a/servo/connectors/kube_metrics.py
+++ b/servo/connectors/kube_metrics.py
@@ -410,8 +410,7 @@ class KubeMetricsConnector(servo.BaseConnector):
                     # Set requests = limits if not specified
                     if (
                         cpu_request := cpu_resources[ResourceRequirement.request]
-                        is None
-                    ):
+                    ) is None:
                         cpu_request = cpu_resources[ResourceRequirement.limit]
 
                     if SupportedKubeMetrics.MAIN_CPU_REQUEST in target_metrics:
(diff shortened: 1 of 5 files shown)


servo/connectors/kube_metrics.py
```diff
@@ -407,80 +407,76 @@ async def periodic_measure(
                     cpu_resources = target_resource_container.get_resource_requirements(
                         "cpu"
                     )
+                    # Set requests = limits if not specified
+                    cpu_request = cpu_resources[ResourceRequirement.request]
+                    if cpu_request is None:
```
linkous8, line 412: Using := is more pythonic btw, you just have to use parenthesis to ensure the order of operations doesn't wind up assigning the boolean expression to your variable instead of the value you're testing
```suggestion
                    if (cpu_request := cpu_resources[ResourceRequirement.request]) is None:
```

servo/connectors/kube_metrics.py
```diff
@@ -514,9 +510,17 @@ async def periodic_measure(
                         "cpu"
                     )
                 )
+                # Set requests = limits if not specified
+                cpu_request = cpu_resources[ResourceRequirement.request]
+                if cpu_request is None:
```
linkous8, line 515: see line 411 comment

linkous8 (approved):
👍 👍 👍

PR #427, opened 2022-05-11
Bugfix/update poetry install process

blakewatters:
Updated the Poetry install process to utilize the new Poetry installation method

### DIFF
--- a/.github/workflows/installation.yaml
+++ b/.github/workflows/installation.yaml
@@ -9,44 +9,44 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-      with:
+      - uses: actions/checkout@v2
+        with:
           fetch-depth: 1
-    - name: Output Python version
-      id: python-version
-      run: |
-        echo "::set-output name=python-version::$(cat .python-version)"
-    - name: Cache pyenv versions
-      uses: actions/cache@v2.1.1
-      id: cached-pyenvs
-      with:
-        path: /opt/hostedtoolcache/pyenv_root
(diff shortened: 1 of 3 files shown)


.github/workflows/installation.yaml
```diff
@@ -9,44 +9,44 @@ jobs:
...
+          key: ${{ runner.os }}-pyenv
+      - name: Install pyenv versions
+        uses: gabrielfalcao/pyenv-action@v9
+        with:
+          default: ${{ steps.python-version.outputs.python-version }}
+        # if: steps.cached-pyenvs.outputs.cache-hit != 'true'
+      - name: Install Poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+          export PATH="$HOME/.local/bin:$PATH"
```
linkous8, line 33:
```suggestion
```

This is already onPATH  (see https://github.com/linkous8/test/runs/6408576299?check_suite_focus=true)

blakewatters: Since I can't actually execute any CI operations right now I'm sorta shadowboxing over here. But here you go, the export statements have been removed from the workflow

linkous8: CI operations come free with public repos. I just created the above repo to verify the path myself since I couldnt find definitive documentation 😉

linkous8:
Any objections to me taking this branch and pushing it up to the opsani org to run the test suite? I'm working on a setup to where that won't be necessary but its up to you what we do in the interim.

blakewatters:
Don't care. Just do something. This is the correct set of changes for where Poetry is going

PR #430, opened 2022-05-11
test docker auth v2

benburdick:


blakewatters:
seems pretty reasonable. why holding?

PR #431, opened 2022-05-12
Ensure mypy execution always returns a 0 status code

blakewatters:


.github/workflows/test.yaml
```diff
@@ -113,8 +113,7 @@ jobs:
       - name: Run mypy
         run: |
           source .venv/bin/activate
-          # NOTE: This will not fully succeed for the time being
-          mypy servo/
+          mypy --pretty servo/ || true
```
linkous8, line 116: Can you come up with a solution other than returning false positive green from a failing test? 

The current plan is to address all typing issues in an upcoming refactor intended to reduce the size of the code base as well as improve readability

linkous8 (changes requested):

blakewatters:
The test suite is broken because you decided to integrate a highly pedantic linter onto the main branch without fixing any of the issues it raised.

My change here provides the `mypy` feedack into CI logs that are actionable and can be remediated at your leisure.

I utilized type hints during development to provide a valuable signal to my future self and other engineers working on the code.

Conforming to linter rules that I did not even know about during development does not mean that the utilization of typing in Python, which is I will remind you totally optional and only enforced by conventions and linting tools, is incorrect or indicates quality issues.

Are there specific bugs and design problems that you are aware of that are being pointed out by this `mypy` execution that cannot and should not be safely ignored to roll up the remaining test outputs into a positive/negative outcome?

And if so, why have you let them ride for 45 days and counting?

Why do you believe that leaving the test suite in an intentionally broken state when it actually works is the correct course of action?

linkous8:
Its a caveat emptor for any readers not to put too much stock in the type hints. They convey the intent well enough but should not be relied on as precise syntax yet. I'd say this interaction is evidence of the test having the desired effect (oiling the wheel). 

> The codebase works right now, full stop, no questions asked but you are forcing it to present itself as broken for no reason

Like you said, typing is optional so anyone with an understanding of these things would know mypy failing isn't and indicator of brokenness

blakewatters:
You are still somehow missing the point.

The software totally works right now.

You are artificially broadcasting a signal that it does not for no reason.

It's absolutely ridiculous and unprofessional.

Why do you want this? What are you achieving?

blakewatters:
Are you high?

You continue to evade the fundamental questions:

- Does the software work?
- Will fixing these type hint linter violations improve utility and/or user experience?
- If it is so important, why have you not do not it?

linkous8:
It probably boils down to stylistic difference here. As I recall, you prefer a decisive "knockout punch" whereas I don't mind a few ugly, misshapen stepping stones along the path (however long that path turns out to be). Just trying to have an honest and open discussion here.

> - Does the software work?
yes
> - Will fixing these type hint linter violations improve utility and/or user experience?
yes
> - If it is so important, why have you not done it?
because the software functions correctly, its hard to prioritize changes that are more or less superficial. Even still, ux is benefited by the honesty about the shortcomings in the current state of the type hints; this is provided by the failing test as well as the vscode extension recommendations that turn on mypy inline linting.

linkous8:
> If the software functions correctly, then how can it affect the UX (User Experience)? Have you confused UX with DX (Developer Experience)?

In this particular product (and my larger, likely controversial, personal view) the line between User and Developer is pretty blurred and only going to blur further as we seek to create an environment where users are encouraged and empowered to solve their own problems. (in my view, thats the whole purpose of open source software)

> Okay, so explain how conforming to these linter violations are going to improve the utility and user experience?

Given there are several major abstractions in the code base (extensive mixin use, dynamic configuration, event bus), we need to ensure the correctness of the typing of the methods and helpers that comprise our API before we can realistically expect external contribution. Anything else is putting the cart before the horse but there isn't even a pressing need to drive said cart yet so 🤷 

> I can be persuaded to immediately contribute my efforts to slay these dragons.

Slay away good sire 🐴 ⚔️

blakewatters:
May have lost the plot here but you are still saying that keeping the test suite intentionally broken for no reason is the move?

linkous8:
Communication is key 🔑

blakewatters:
Did we get DJ Khaled involved or are we still writing software professionally?

linkous8:
Superseded by #463

PR #439, opened 2022-06-29
Optserv 27 stateful set support for kubernetes connector of servox

linkous8:
Trigger integration tests and open floor for design discussion. Note this is still a draft


mjdeclerck (commented):
This is probably a big enough change that I am not qualified to review.  I would swap me out for Peter.

rstarmer (approved):
@linkous8 this looks good, at some point we should break this up into smaller units (is that even feasible with the craziness of K8s?!?!).

LGTM

PR #446, opened 2022-08-09
Optserv 90 resolve servox dependabot security alerts

linkous8:


mjdeclerck (approved):
LGTM - All of these changes seem sane on the surface.  The only one that triggers any comment is the "await asyncio.sleep(1)".  I have a general aversion to sleeps in code but it seems relatively inoculous.

linkous8:
> The only one that triggers any comment is the "await asyncio.sleep(1)". I have a general aversion to sleeps in code but it seems relatively innocuous.

Thats' a fair critique. In this case, its meant to add some buffer to prevent the API server from being thrashed by repeated requests. Any alternative proposals are more than welcome

mjdeclerck:
I understand what it is doing and why, so I am fine with it.  The reason I flagged it is that I have seen a pattern during my career of sleep functions being used to work around issues rather than understanding/solving them.  In this case, it seems fine.  If I saw the timeout increasing from 1 -> 5 -> 10 over time, that would likely indicate it was masking and not solving the problem.

PR #451, opened 2022-08-23
Update version and changelog for v2.1.1

linkous8:


linkous8:
NOTE: required checks are being bypassed as this PR is purely documentational. Further, the same code is tested in the following run and passes https://github.com/opsani/servox/actions/runs/2913607994

PR #475, opened 2022-11-01
Change envoy image tag to latest upstream

linkous8:


linkous8:
Bypassing checks to to intermittent auth issues

PR #499, opened 2023-02-13
Optserv 482 update servox to work with agent management sidecar

linkous8:
- Update AppDynamics optimizer to read configuration from a local file set by configmap
  - File will be watched for changes and reload the file when they occur
- Update helm chart to add the sidecar connection setting type which will configure the agent management sidecar and setup the volume and mount for the connection file it manages (as well as passing it into servo's configuration)


helm/templates/deployment.yaml
```diff
@@ -66,18 +66,73 @@ spec:
...
+      {{- if eq .Values.connectionSettingsType "sidecar" }}
+      - name: agent-management-sidecar
+        image: {{ required ".Values.agentMgmtSidecarImage is required for sidecar connection settings type" .Values.agentMgmtSidecarImage }}
+        resources:
+          limits:
+            cpu: 100m
+            memory: 128Mi
+        env:
+        - name: SERVICE_URL
+          value: {{ default "<SERVICE_URL>" .Values.agentMgmtServiceUrlOverride }}
```
pavel-georgiev, line 86: Is this an acceptable default? (same for other uses of default "<...>")

pavel-georgiev: Separate note on naming - I see variables named `somethingOverride`, can we rename them to just `something` and keep them optional.

helm/templates/secret-sidecar.yaml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2023 Cisco Systems, Inc. and/or its affiliates.
```
pavel-georgiev, line 1: Suggestion: Since this sets the same secret name as the one from secret.yaml, consider putting in in the same file (makes it clearer that it is the same object, but the content is conditional on the connection type). Also, some of the secret data is common (metadata, etc.) so we will avoid some duplication.

pavel-georgiev (changes requested):
I have added some inline suggestions, mostly on naming.

In addition, I have two more requests:
- Can we run a long test of the servo (1000+steps), probably with the emulator or any other connector that can quickly iterate through steps. I would like to validate that some of the new changes (mostly the filesystem watch) work well and have no leaks that only show up after sustained usage
- It would make sense to have servo be tolerant to unexpected output by the sidecar (if that is not the case already). For example - if the sidecar writes something in the shared file that we fail to parse and we are in the middle of an operation, I would rather log the error and proceed as if the fie has not changed. If the problem was temporary, then we will pick up the next change. If not, our credentials will expire eventually and we will get the same result (unable to send result), but (a) we have a chance to recover from temp errors and (b) we are not subject to sudden interruptions caused by the sidecar (i.e. servo crash caused by sidecar). Few scenarios to consider: (1) sidecar writes yaml that is invalid/incomplete, (2) sidecar does not update credentials (we can't overcome this, but we should make sure that command is not interrupted because of this), (3) sidecar write wrong credentials that can be parsed, but do not pass authentication

helm/templates/deployment.yaml
linkous8: >Is this an acceptable default? (same for other uses of default "<...>")

This is the format expected by the orchestrator in order to perform substitution. Here's the slack thread for reference https://appdynamics.slack.com/archives/C0468BYM85C/p1675881169951929?thread_ts=1675877822.302409&cid=C0468BYM85C

>I see variables named somethingOverride

This makes it clear the end user is not expected to provide these values since the whole point of this configuration is to have them populated externally.

PR #503, opened 2023-02-22
Optserv 428 create servo no op measure driver for orbit release

linkous8:
Add WaitConnector which sends a heartbeat for the Control specified duration which external measurement is being conducted

### DIFF
--- a/servo/connectors/wait.py
+++ b/servo/connectors/wait.py
@@ -1,0 +1,53 @@
+import servo
+
+
+class WaitConfiguration(servo.BaseConfiguration):
+    measure_enabled: bool
+
+
+@servo.metadata(
+    description="Wait (no-op) Connector for Opsani",
+    version="0.0.1",
+    homepage="https://github.com/opsani/servox",
+    license=servo.License.apache2,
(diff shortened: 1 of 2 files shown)


servo/connectors/wait.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @servo.on_event()
+    def describe(self, control: servo.Control = servo.Control()) -> servo.Description:
+        return servo.Description(metrics=[])
+
+    @servo.on_event()
+    async def measure(
+        self, *, metrics: list[str] = None, control: servo.Control = servo.Control()
+    ) -> servo.Measurement:
+        if self.config.measure_enabled:
```
pavel-georgiev, line 50: Does that mean that a configuration is required for this driver? If yes - what is the motivation behind requiring configuration for this? 

Even if we have a use case for measure_enabled=False, we should probably make the default to match the most common use case.

linkous8: Resolved offline. (Main reason I advocated for it is to improve readability of the configuration file since some form of config file entry is required to have the connector included in the assembly)

PR #562, opened 2023-12-13
Fix servo run_servo being called twice preventing clean shutdown

linkous8:
- bump virtualenv to fix local runs of pre-commit


pavel-georgiev (commented):
Thanks for the quick turnaround. Can you please test before/after the change to make sure that the behavior described in the bug report is observable before the change and is fixed after the change

PR #571, opened 2024-04-26
Optserv 1269 part 2 servox async logic hardening

linkous8:
- refactor connector initialization to be member method load_connectors of servo instead of procedural assembly logic
- update servo to store connector type routes loaded from config yam/default values
- fix servo attach() and detach() referencing servo module
- refactor servo shutdown to send gootbye event, add reason param to shutdown to include in request
- use TaskGroup for checks rememdy processing to avoid task leaks
- Move AssemblyRunner run() args poll, interactive to class properties set on init
- fix incorrected type hint on _default_context_manager
- Update progress handler to support creating its queue processor task with an external task group tied to lifetime of the AssemblyRunner
- Update progress handler sink to require servo from extra or context and include in progress dict
- alphebetize progress dict by key
- move servo to api connection details from ServoRunner to servo
- Update ServoRunner to utilize a task group to organize tasks that used to be created and stored as properties with additional management boilerplate
- Coallesce main_loop into run_main_loop
- get rid of pointlress _reraise_if_neccessary method cluttering stack trace
- remove hello event logic from run and set it to run in servo.startup()
- fix main_loop_task not set to main loop of ServoRunner
- move goodbye event logic to servo shutdown, call assembly.remove_servo() to send it, use _main_loop_task and _task_group for cleanup and auditing
- Update AssemblyRunner to kick off and store single root task from synchronous invocation
- Move most of run() logic into async version _run()
- Update AssemblyRunner to use two task group for tracking tasks that used to be spawned from create_task without storing any refernces to them
- _task_group contextualizes all tasks run within the AssemblyRunner._run() function to ensure the lifetime of the Progress handler is tied to the lifetime of the running servos and gets cancelled by crash errors thrown by them
- _runners_task_group holds only the servo tasks. Due to the progressHandlers queue_processor running an infinite loop, _task_group would run forever if allowed to. _runners_task_group allows us to determine if all servo runner tasks have completed and gracefully exit the _task_group when that is the case
- move _report_progress from dynamic method to class method
- move _handle_progress_exception from dynamic method to class method
- add AssemblyRunner.shutdown guard for re-entrant shutdown
- Update AssemblyRunner.shutdown to attempt to shutdown gracefully and kill the root_task instead of killing all tasks. This should propagate a cancellation down the chain of task groups this agent is now structured under
- Update _handle_progress_exception to shutdown and restart only the individual servo that threw the error instead of all_tasks. Makes use of previously dead code Assembly.add/remove_servo as well as added task groups
- Update Servo connectors lists init to be self contained
- Fix type hint error warnings caused by DNSSubdomainName, DNSLabelName, ContainerTagName
- Disable unused pubsub startup in prometheus connector
- fix unresolved import errors in IDE for devtools.debug
- Fix captured_logs type hint error
- Fix type hints for minikube fixture
- Use task group for FakeAPI server
- Use task group for kubectl_ports_forwarded
- Remove any remaining logic that clobbers asyncio.all_tasks (only tests left at this point)
- fix servo_test failures
- replace prometheus test tasks with task group

### DIFF
--- a/servo/api.py
+++ b/servo/api.py
@@ -120,8 +120,12 @@ class Status(pydantic.BaseModel):
         return cls(status=ServoStatuses.ok, message=message, reason=reason, **kwargs)
 
     @classmethod
-    def from_error(cls, error: servo.errors.BaseError, **kwargs) -> "Status":
-        """Return a status object representation from the given error."""
+    def from_error(
+        cls, error: servo.errors.BaseError | ExceptionGroup, **kwargs
+    ) -> "Status":
+        """Return a status object representation from the given error (first if multiple in group)."""
+        if isinstance(error, ExceptionGroup):
+            error = error.exceptions[0]
         if isinstance(error, servo.errors.AdjustmentRejectedError):
             status = ServoStatuses.rejected
         elif isinstance(error, servo.errors.EventAbortedError):
--- a/servo/api.py
+++ b/servo/api.py
@@ -54,6 +54,16 @@ class ServoStatuses(enum.StrEnum):
     aborted = "aborted"
     cancelled = "cancelled"
 
+    def from_error(error: Exception) -> ServoStatuses:
+        if isinstance(error, servo.errors.AdjustmentRejectedError):
+            return ServoStatuses.rejected
+        elif isinstance(error, servo.errors.EventAbortedError):
+            return ServoStatuses.aborted
+        elif isinstance(error, servo.errors.EventCancelledError):
+            return ServoStatuses.cancelled
+        else:
+            return ServoStatuses.failed
+
 
 Statuses = Union[OptimizerStatuses, ServoStatuses]
 
@@ -127,38 +137,28 @@ class Status(pydantic.BaseModel):
         cls, error: servo.errors.BaseError | ExceptionGroup, **kwargs
     ) -> "Status":
         """Return a status object representation from the given error (first if multiple in group)."""
-        reason = getattr(error, "reason", ...)
         if isinstance(error, ExceptionGroup):
             servo.logger.warning(
                 f"from_error executed on exceptiongroup {error}. May produce undefined behavior"
             )
             status = ServoStatuses.failed
             try:
-                main_err = servo.errors.ServoError.servo_error_from_group(
+                error = servo.errors.ServoError.servo_error_from_group(
                     exception_group=error
                 )
-                if isinstance(main_err, list):
-                    main_err = main_err[0]
-                reason = main_err.reason
-            except:
+                if error._additional_errors:
+                    additional_messages = [str(e) for e in error._additional_errors]
+                    kwargs["additional_messages"] = (
+                        kwargs.get("additional_messages", []) + additional_messages
+                    )
+            except Exception:
                 servo.logger.exception(
                     "Failed to derive exceptiongroup reason for api response"
                 )
                 pass
-        if isinstance(error, servo.errors.AdjustmentRejectedError):
-            status = ServoStatuses.rejected
-        elif isinstance(error, servo.errors.EventAbortedError):
-            status = ServoStatuses.aborted
-        elif isinstance(error, servo.errors.EventCancelledError):
-            status = ServoStatuses.cancelled
-        else:
-            status = ServoStatuses.failed
 
-        if error._additional_errors:
-            additional_messages = [str(e) for e in error._additional_errors]
-            kwargs["additional_messages"] = (
-                kwargs.get("additional_messages", []) + additional_messages
-            )
+        reason = getattr(error, "reason", ...)
+        status = ServoStatuses.from_error(error)
 
         return cls(status=status, message=str(error), reason=reason, **kwargs)
 
--- a/servo/api.py
+++ b/servo/api.py
@@ -105,6 +105,9 @@ class Request(pydantic.BaseModel):
 class Status(pydantic.BaseModel):
     status: Statuses
     message: Optional[str] = None
+    other_messages: Optional[
+        list[str]
+    ] = None  # other lower priority error in exception group
     reason: Optional[str] = None
     state: Optional[Dict[str, Any]] = None
     descriptor: Optional[Dict[str, Any]] = None
@@ -124,8 +127,24 @@ class Status(pydantic.BaseModel):
         cls, error: servo.errors.BaseError | ExceptionGroup, **kwargs
     ) -> "Status":
         """Return a status object representation from the given error (first if multiple in group)."""
+        reason = getattr(error, "reason", ...)
         if isinstance(error, ExceptionGroup):
-            error = error.exceptions[0]
+            servo.logger.warning(
+                f"from_error executed on exceptiongroup {error}. May produce undefined behavior"
+            )
+            status = ServoStatuses.failed
+            try:
+                main_err = servo.errors.ServoError.servo_error_from_group(
+                    exception_group=error
+                )
+                if isinstance(main_err, list):
+                    main_err = main_err[0]
+                reason = main_err.reason
+            except:
+                servo.logger.exception(
+                    "Failed to derive exceptiongroup reason for api response"
+                )
+                pass
         if isinstance(error, servo.errors.AdjustmentRejectedError):
             status = ServoStatuses.rejected
         elif isinstance(error, servo.errors.EventAbortedError):
@@ -135,7 +154,13 @@ class Status(pydantic.BaseModel):
         else:
             status = ServoStatuses.failed
 
-        return cls(status=status, message=str(error), reason=error.reason, **kwargs)
+        if error._additional_errors:
+            additional_messages = [str(e) for e in error._additional_errors]
+            kwargs["additional_messages"] = (
+                kwargs.get("additional_messages", []) + additional_messages
+            )
+
+        return cls(status=status, message=str(error), reason=reason, **kwargs)
 
     def dict(
         self,
--- a/servo/assembly.py
+++ b/servo/assembly.py
@@ -244,12 +244,12 @@ class Assembly(pydantic.BaseModel):
             )
         )
 
-    async def shutdown(self):
+    async def shutdown(self, reason: str | None = None):
         """Notify all servos that the assembly is shutting down."""
         await asyncio.gather(
             *list(
                 map(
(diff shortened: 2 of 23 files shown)


servo/api.py
```diff
@@ -120,8 +123,28 @@ def ok(
...
+    def from_error(
+        cls, error: servo.errors.BaseError | ExceptionGroup, **kwargs
+    ) -> "Status":
+        """Return a status object representation from the given error (first if multiple in group)."""
+        reason = getattr(error, "reason", ...)
+        if isinstance(error, ExceptionGroup):
+            servo.logger.warning(
+                f"from_error executed on exceptiongroup {error}. May produce undefined behavior"
+            )
+            status = ServoStatuses.failed
```
pavel-georgiev, line 135: Would that mask more specific exceptions?

servo/checks.py
```diff
@@ -757,53 +757,52 @@ async def process_checks(
...
+
+                                    async def awaitable() -> None:
+                                        result = failure.remedy()
+                                        if asyncio.iscoroutine(result):
+                                            await result
+
+                                    _ = tg.create_task(awaitable)
+                                    servo.logger.info("💡 Attempting to apply remedy...")
+
+                        except asyncio.TimeoutError:
```
pavel-georgiev, line 801: Is that supposed to catch the expiation from the previous except block?

It looks like it should be:
```
                        try:
                            await asyncio.wait_for(task, 10.0)
                        except asyncio.TimeoutError as error:
                            try:
                                async with asyncio.timeout(10.0):
                                    ... # remedy
                            except asyncio.TimeoutError:
                                    ... # remedy timeout out
```

Also - do we need to catch other exceptions here as well?

servo/connectors/kubernetes.py
```diff
@@ -2249,8 +2278,13 @@ async def detach(self, servo_: servo.Servo) -> None:
     async def describe(
         self, control: servo.Control = servo.Control()
     ) -> servo.Description:
-        state = await self._create_optimizations()
-        return state.to_description()
+        try:
+            state = await self._create_optimizations()
+            return state.to_description()
+        except ExceptionGroup as eg:
```
pavel-georgiev, line 2284: Would that catch it if a single exception is raised?

pavel-georgiev (approved):

servo/servo.py
```diff
@@ -233,19 +238,26 @@ async def dispatch_event(
...
 
     def __init__(
-        self, *args, connectors: list[servo.connector.BaseConnector], **kwargs
+        self,
+        *args,
+        connectors: list[servo.connector.BaseConnector] | None = None,
+        __connectors__: list[servo.connector.BaseConnector] | None = None,
+        **kwargs,
     ) -> None:  # noqa: D107
-        super().__init__(*args, connectors=[], **kwargs)
```
leokalev, line 238: The use of 'connectors' vs '__connectors__' is not easy to understand by looking at the code. If I got it right, the __connectors__ is exclusively for the 'events' mixin and is a copy of 'connectors', except it also includes 'self'.

A short comment explaining the use might be helpful here.

servo/servo.py
```diff
@@ -435,6 +479,50 @@ def top_level_schema_json(self, *, all: bool = False) -> str:
...
         )
 
+    async def connect(self) -> None:
+        # must define dynamic function to allow configurable backoff
+        await backoff.on_exception(
+            backoff.expo,
+            httpx.HTTPError,
+            max_time=lambda: self.config.settings.backoff.max_time(),
+            max_tries=lambda: self.config.settings.backoff.max_tries(),
+            on_giveup=lambda details: self.logger.critical(
```
leokalev, line 489: max_tries=lambda: self.config.settings.backoff.max_tries()

should be equivalent to:
 max_tries=self.config.settings.backoff.max_tries
 
 (there doesn't seem to be any need to define a function that just calls another function with the same exact signature - might as well just pass the reference to the latter)

servo/servo.py
```diff
@@ -435,6 +479,50 @@ def top_level_schema_json(self, *, all: bool = False) -> str:
...
+        try:
+            await self.post_event(
+                servo.api.Events.hello,
+                dict(
+                    agent=servo.api.user_agent(),
+                    telemetry=self.telemetry.values,
+                ),
+            )
+        except httpx.HTTPStatusError as error:
+            if (
```
leokalev, line 505: Given that you have a backoff() decorator  on this, it may be worth it configuring the backoff not to retry on 410
(this is cosmetic, won't affect operation very much - add the exclusion only if it is easy to do so)

leokalev (changes requested):

servo/servo.py
```diff
@@ -435,6 +479,50 @@ def top_level_schema_json(self, *, all: bool = False) -> str:
...
+            max_time=lambda: self.config.settings.backoff.max_time(),
+            max_tries=lambda: self.config.settings.backoff.max_tries(),
+            on_giveup=lambda details: self.logger.critical(
+                f"retries exhausted after {details['elapsed']} seconds, giving up"
+            ),
+        )(self.try_connect)()
+
+    async def try_connect(self) -> None:
+        self.logger.info("Saying HELLO.", end=" ")
+        try:
```
leokalev, line 496: what is the end=" " for?
(the log message doesn't seem have a replacement string that looks like '{end}' to be substituted for this kw-arg)

(if this were python's built-in print(), it might have an effect - though not a desirable one).

leokalev:
note: (follow up on the comment about changing 'black' to a larger line_length):
if you decide to adopt this, please don't do it as part of the current PR - that would be messy >:-0

servo/api.py
```diff
@@ -120,8 +123,28 @@ def ok(
...
+                    exception_group=error
+                )
+                if isinstance(main_err, list):
+                    main_err = main_err[0]
+                reason = main_err.reason
+            except:
+                servo.logger.exception(
+                    "Failed to derive exceptiongroup reason for api response"
+                )
+                pass
```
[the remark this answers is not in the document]
linkous8: > Please use 'except Exception' (or, if at all possible, specify which exception(s) you want to catch).
> remove 'pass'

Will do

> (note, 'reason' will remain set to "...", is this ok for the subsequent cls(..., reason=reason)?

Its preferable for reason to be ... since pydantic treats that as an unset field instead of us setting it to null which would be included on a json dump.

servo/assembly.py
```diff
@@ -224,12 +209,14 @@ async def add_servo(self, servo_: servo.servo.Servo) -> None:
...
+        await servo_.attach()
 
         if self.is_running:
-            await servo.startup()
+            await servo_.startup()
 
-    async def remove_servo(self, servo_: servo.servo.Servo) -> None:
+    async def remove_servo(
+        self, servo_: servo.servo.Servo, reason: str | None = None
+    ) -> None:
```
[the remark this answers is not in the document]
linkous8: Thanks for the feedback, I will do that in a future PR to avoid crowding the diff on this PR any further

servo/events.py
```diff
@@ -364,7 +365,7 @@ def create_event(
 
     def _default_context_manager() -> AsyncContextManager:
         # Simply yield to the on event handler
```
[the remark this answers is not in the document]
linkous8: The 'on' bit refers to the event preposition https://github.com/opsani/servox?tab=readme-ov-file#event-handlers--prepositions

servo/servo.py
```diff
@@ -219,12 +219,17 @@ class Servo(servo.connector.BaseConnector):
     """The active connectors in the Servo.
     """
 
+    routes: dict[str, type[servo.connector.BaseConnector]] = {}
+    """The dict of connectors this servo was initially configured with. Used for adding connectors that have been removed
+    """
```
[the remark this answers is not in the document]
linkous8: If CO requests a command cancellation, the servo is restarted in order to kill the currently executing command. This causes the connectors to be removed as they are shutdown and detached to fulfill the event lifecycle (note this may be refactored to only kill tasks rather than shutdown servo altogether but the current form maintains alignment with the previous implementation)

https://github.com/opsani/servox/pull/571/files#diff-7ed9452d2ccb04684818b97a667e535afc48f1ccd8b6f8c903c130d00d94e79bR772-R779

servo/servo.py
linkous8: I don't understand it 100% either but I think one is used for the event bus (includes servo) for dispatching events and the other is used by the servo (excludes servo self reference) for non-event collector handling

servo/servo.py
linkous8: The lambda encloses a reference to the original servo so if the config is altered dynamically the updated config will be applied instead of the decorator using a stale method of a possibly garbage collected config object (that said this level of dynamicness is not currently in use)

servo/servo.py
linkous8: good catch, old protoype cruft https://github.com/opsani/servox/commit/f63f85fc2d6b216f1bb2f5ee38087db5ba958dae#diff-7f314f2efdab903b2e4d8a3ea73e35e671b8a3bd3b344af7b2977b636e99598eL106

servo/servo.py
linkous8: Deferring this work for a future PR

servo/servo.py
leokalev: The markdown let me down here - I meant `__connectors__` not __connectors__. Anyway, this is optional - but if you have an idea what to put in a docstring, it will be great.

servo/servo.py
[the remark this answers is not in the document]
leokalev: Maybe I misunderstood the code, but it seems that 'routes' is used for the initial setup, as well (not only when re-connecting it after it is restarted). That's why I found the comment confusing.

If I got it wrong and this is for re-start only, then maybe: "used for restoring this servo to its initial configuration after it is re-started"

If it is actually used both for start and re-start: "used for adding the connectors to this servo on initialization and on re-start"

servo/servo.py
linkous8: It should go in the events mixin but really there should be some logic to unify the variable so that its less confusing and the mixin is self-contained instead of having keyword arg details spilling out into other components.

https://github.com/opsani/servox/blob/f3ade34ce015fa64271f12d62518c0777d88ae81/servo/events.py#L646

leokalev (approved):

